PR #7, opened 2016-02-21
Add Swagger definition of the API

malor:


tox.ini
```diff
@@ -15,3 +16,5 @@ commands =
   flake8 xsnippet_api/ tests/
 # check that unittests are passed
   py.test --cov xsnippet_api/ --cov-append tests/
+# check the swagger definition of the API
+  swagger-flex -s contrib/swagger.yaml
```
ikalnitsky: Hm.. The reason I put `setup.py check` and `flake8` check under `[testenv]` declaration is that we usually want to perform those checks for each Python interpreter. Why? Because the results may be different in those cases (`setup.py` may have if branches, and `flake8`, if I'm not mistaken, may work differently for each interpreter).

What about `swagger-flex`? Does it performs interpreter-independent check? If so, I'd prefer to move it into separate `tox` env. What do you think?

malor: It's still a ``tox`` virtual environment, i.e. the ``xsnippet_api`` package will be installed before running the ``commands`` section from tox.ini, thus, all dependencies specified in ``setup.py`` are also installed and some of them (e.g. ``aiohttp``) are python3.4+ only.

I'm fine with moving ``swagger-flex`` into a separate ``testvenv`` and even support it. It's just I'll have to specify the ``basepython = python3.5`` for that env.

ikalnitsky: > It's still a tox virtual environment, i.e. the xsnippet_api package will be installed before running the commands section from tox.ini

We can skip installation for `swagger`, since we don't need it. Example:

```
[testenv:swagger]
skip_install = true
deps =
  swagger-flex
commands =
  swagger-flex -s contrib/swagger.yaml
```

> I'm fine with moving swagger-flex into a separate testvenv and even support it. It's just I'll have to specify the basepython = python3.5 for that env.

I didn't get that. For what reason we need to use `basepython = python3.5`? In case of `tox` is run under `python2`?

ikalnitsky:
@malor, one more thing: could you please add `contrib/` folder to `MANIFESTS.in`? So it's will be a part of release dist.

tox.ini
malor: We only need it because the package and all its dependencies are installed into a virtual environment before running any commands.

I didn't know about ``skip_install = True`` - that will help, thanks!

contrib/swagger.yaml
```diff
@@ -0,0 +1,439 @@
...
+    get:
+      summary: Retrieve information about code snippets.
+      description: |
+        The Snippets endpoint returns information about stored code snippets. The
+        listed snippets are sorted by the date-time of creation in the ascending order.
+
+        By default, only `limit` latest snippets are returned. Paginated queries
+        should be used in order to get more snippets.
+      parameters:
+        - name: filetype
```
ikalnitsky: I'm ok with `filetype`, however, maybe it makes sense to sense to rename it into `syntax`? What do you think?

contrib/swagger.yaml
```diff
@@ -0,0 +1,439 @@
...
+          type: integer
+          minimum: 1
+          format: int64
+          required: true
+          in: path
+          description: Unique identifier representing the specific code snippet.
+        - name: content
+          in: body
+          schema:
+            $ref: '#/definitions/Changeset'
```
ikalnitsky: Hm.. So `PATCH` on `/snippets/:id` must contain only 'changeset' body, correct?

I don't know what's better, but usually RESTful API says:

* use `PUT` when you need to replace resource content 'As Is' (in our case, snippet content, with implicit creation of new changeset)
* use `PATCH` when you need to replace only passed part of resource (in our case, snippet content, with, again, implict creation of new changeset in case `content` has been changed)

So we need to add `PUT` if we want to support, for instance, title renaming or something. That makes sense for `PATCH` either (if we want to have title renaming without creating a new changeset).

What do you think? I just a little bit confused.

contrib/swagger.yaml
malor: I don't really mind, ``syntax`` sounds fine. I wasn't comfortable with ``language``.

contrib/swagger.yaml
malor: I was going to say only new ``changesets`` can be created, but it makes sense to support update for ``title`` and ``tags`` fields at least too.

At the same time, change of ``author_id`` must be prohibited. I'm not sure about change of ``syntax``: it's currently set on a ``snippet``, not ``changeset``, i.e. we should prohibit changes of both independently, otherwise it's possible for them to be not aligned.

contrib/swagger.yaml
```diff
@@ -0,0 +1,439 @@
...
+          schema:
+            $ref: '#/definitions/Snippet'
+          required: true
+          description: |
+              A snippet object, which will be used for update. Can have all or
+              only some of the fields set. Only set fields will be used for
+              update. If new ``content`` is provided, a new changeset of the
+              snippet will be created and set as the current one.
+      responses:
+        201:
```
ikalnitsky: Hm.. `201` means `CREATED`. Do we really want to return that status? What do you think about '200 OK' ?

contrib/swagger.yaml
```diff
@@ -0,0 +1,439 @@
...
+
+definitions:
+  Syntax:
+    type: string
+    enum: *ALLOWED_SYNTAXES
+
+  Changeset:
+    type: object
+    properties:
+      revision:
```
ikalnitsky: Do we really need to have `revision` fixed on data model layer? I thought if it's a list of changesets we can consider its index as a revision number. What do you think?

contrib/swagger.yaml
```diff
@@ -0,0 +1,439 @@
...
+      description: Retrieve all changesets of the snippet.
+      parameters:
+        - name: id
+          type: integer
+          minimum: 1
+          format: int64
+          required: true
+          in: path
+          description: Unique identifier representing the specific code snippet.
+      responses:
```
ikalnitsky: I think `401`, `403` and `404` are also applied. Or you want to hide them and use `400` in all cases?

contrib/swagger.yaml
malor: It's what left after the previous version of ``PATCH`` that only allowed to create new changesets. ``200`` seems to be the correct code to return on update of the resource. Fixed. Thanks!

contrib/swagger.yaml
malor: Not that I'm against it, but then we should consider adding a ``HEAD`` action for  ``/snippets/{id}/changesets``, so that it's possible to retrieve the current changeset index without downloading all changesets content (what ``GET`` what do).

contrib/swagger.yaml
malor: Good catch!  Thanks!

PR #9, opened 2016-02-25
Add testing facilities

ikalnitsky:
So far asyncio is quite new framework and, therefore, has lack of
testing instruments. This commit introduces convenient testing
facilities such as:

* inloop decorator

  .. that runs a wrapped function in newly created event loop.

* AIOTestMeta metaclass

  .. that runs classes' async methods in newly created event loop.

* AIOTestApp class

  .. that is a facade to test an aiohttp.web.Application instances.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,0 +1,142 @@
+"""
+    tests
+    -----
+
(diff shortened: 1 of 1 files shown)


malor:
lgtm

PR #13, opened 2016-02-28
tests: separate init and cleanup steps from test cases

malor:
Move the repeating initialization (e.g. app creation) and clean up
steps (e.g. DB cleanup) into setup() / teardown() method helpers,
which will be executed before and after each test case.

setup() / teardown() must be aware of the event loop instance which
was created for a particular test case, so that they can execute
appropriate actions (e.g. clean up a DB), thus, setup() / teardown()
are executed by the inloop() decorator.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -16,7 +16,7 @@ import inspect
 import aiohttp
 
 
-def inloop(fn):
+def inloop(fn=None, *, setup=None, teardown=None):
     """Decorator that runs a wrapped function in the event loop.
 
     Usage::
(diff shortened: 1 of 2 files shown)


tests/__init__.py
```diff
@@ -78,12 +105,31 @@ def test_b(self):
...
+    A test class may contain special setup / teardown methods, which will
+    be executed before and after running of each test case::
+
+        class TestSnippets(unittest.TestCase, metaclass=AIOTestMeta):
+
+            def setup(self):
+                self.app = create_app(DEFAULT_CONF)
+
+            def teardown(self):
+                loop = asyncio.get_event_loop()
```
ikalnitsky: Hm.. What do you think about making it more event-loop agnostic?

It seems we can:

* Check in decorator whether `setup`/`teardown` is a coroutine function.
* If so, run them in newly created event loop.
* If not, run them as regular methods.

In that case it could be rewritten into something like:

```python
def setup(self):
    self.app = create_app(DEFAULT_CONF)

async def teardown(self):
    await self.app['db'].snippets.drop()
```

What do you think?

tests/__init__.py
```diff
@@ -26,37 +26,64 @@ def inloop(fn):
...
+            asyncio.set_event_loop(tmp_loop)
 
-            # running a given coroutine may produce another ones to free
-            # resources (e.g. close sockets) when it's done, so we need
-            # to run the event loop one more time to execute them
-            tmp_loop.stop()
-            tmp_loop.run_forever()
-            tmp_loop.close()
+            if setup:
+                setup(self)
```
ikalnitsky: Does it tie this decorator to be used **only** with test methods? No test functions?

tests/__init__.py
malor: Sure, np. Will update the patch.

tests/__init__.py
malor: As is, yes, due to the fact we have to pass the ``self`` argument if ``setup`` is a method of a class.

Would you prefer to have an inspect call here to analyze if a single ``self`` argument is expected? (I believe, in Python 3.x there is *no* notion of an unbound method - it will appear as a general function to the outside world).


```python

>>> class Spam:
...     def eggs(self):
...         pass
... 
>>> Spam.eggs
<function Spam.eggs at 0x10c73b378>

>>> def eggs(self):
...     pass
... 
>>> inspect.ismethod(Spam.eggs)
False
>>> inspect.ismethod(eggs)
False
>>> inspect.isfunction(eggs)
True
>>> inspect.isfunction(Spam.eggs)
True

```

tests/__init__.py
ikalnitsky: Thank you. Not it's good... at least API's good, the implementation makes me a bit scary! ;)

tests/__init__.py
ikalnitsky: Nah.. let's go with what we have for now.

PR #15, opened 2016-02-28
Add 'pool-size' to default.conf

ikalnitsky:
It's very convenient to setup such options like 'pool-size' through
configuration file, since one might want to turn it taking into
account server resources.


zoresvit:
:+1: Though I don't get why the coverage decreased here :)

PR #16, opened 2016-02-28
Do not pass down event loop, use default one

ikalnitsky:
Having possibility to pass down custom event loop looks like unnecessary
complication that requires ugly workaround (see one with Motor). Let's
remove it, and may the code a little more event-loop agnostic.


malor:
lgtm, let's give the default loop a try

PR #21, opened 2016-03-19
Services

ikalnitsky:


xsnippet_api/services/snippet.py
```diff
@@ -0,0 +1,63 @@
...
+
+    async def create(self, snippet):
+        snippet = self._normalize(snippet)
+        await self.db.snippets.insert(snippet)
+        snippet['id'] = snippet.pop('_id')
+        return snippet
+
+    async def get(self, *, limit=100):
+        snippets = await self.db.snippets.find()\
+            .sort('_id', pymongo.DESCENDING)\
```
malor: Sorting by ``_id`` is a side-effect of having auto-incrementing IDs, which may or may not change in the future :P 

Maybe just sort by ``created_at``?

xsnippet_api/services/snippet.py
```diff
@@ -0,0 +1,63 @@
...
+            .limit(limit)\
+            .to_list(None)
+
+        for snippet in snippets:
+            snippet['id'] = snippet.pop('_id')
```
malor: Was going to suggest to use a list comprehension here, but giving the fact we actually change the source snippets, an ``for`` is actually better.

xsnippet_api/services/snippet.py
```diff
@@ -0,0 +1,63 @@
...
+        rv = dict({
+            'title': None,
+            'author_id': None,
+            'syntax': 'text',
+            'is_public': True,
+            'tags': [],
+        }, **snippet)
+
+        rv['created_at'] = datetime.datetime.utcnow().replace(microsecond=0)
+        rv['updated_at'] = rv['created_at']
```
malor: I think there are at least two strategies of initializing ``updated_at`` field:

1) to set the ``updated_at`` value to be equal to ``created_at`` value on creation of a new entity

2) set it to ``None``

In OpenStack we use the latter. I don't really mind, though: it's still possible to find which snippets have never been updated by looking for snippets with ``updated_at`` == ``created_at``, although there might be an edge case, when a snippet is updated *really* fast after it was created.

tests/resources/test_snippets.py
```diff
@@ -19,73 +19,95 @@
...
-            pkg_resources.resource_filename('xsnippet_api', 'default.conf')
-        )
+            pkg_resources.resource_filename('xsnippet_api', 'default.conf'))
 
         self.app = create_app(conf)
 
     async def teardown(self):
         await self.app['db'].snippets.remove()
 
+    def _snippets(self, id=None):
```
malor: maybe rename the argument name to ``with_ids`` or ``enumerate``?

xsnippet_api/resources/resource.py
```diff
@@ -11,12 +11,29 @@
...
+    Extra support of the following data types:
+
+      - :class:`datetime.datetime` -- encoded into ISO-8601 formatted string
+      - :class:`datetime.date` -- encoded into ISO-8601 formatted string
+    """
+
+    def default(self, instance):
+        if isinstance(instance, (datetime.datetime, datetime.date)):
+            return instance.isoformat()
+        return super(_JSONEncoder, self).default(instance)
```
malor: it's time we started to use super() without arguments :P

xsnippet_api/services/snippet.py
ikalnitsky: Make sense. Thanks, I'll fix it in next patchset.

xsnippet_api/services/snippet.py
ikalnitsky: Hm.. I've checked GitHub and Twitter API, and when they create a new resource both `created_at` and `updated_at` are properly filled and have the same value. I propose to skip it now, and maybe back to that question a bit later.

xsnippet_api/resources/resource.py
```diff
@@ -11,12 +11,29 @@
...
+    """Advanced JSON encoder for extra data types.
+
+    Extra support of the following data types:
+
+      - :class:`datetime.datetime` -- encoded into ISO-8601 formatted string
+      - :class:`datetime.date` -- encoded into ISO-8601 formatted string
+    """
+
+    def default(self, instance):
+        if isinstance(instance, (datetime.datetime, datetime.date)):
```
malor: we should not see the latter, but it's fine to have both

PR #22, opened 2016-03-19
Add a basic Dockerfile

malor:
To build an image run:

  $ docker build -t xsnippet_api .

To run a new container using the created image do:

  $ docker run xsnippet_api

XSNIPPET_API_SETTINGS env variable can be used to specify the path to the
service configuration file (one is expected to mount the configuration into
a container first).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+.gitignore
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
-sudo: false
+dist: trusty
+sudo: required
 language: python
 
 matrix:
@@ -10,13 +11,20 @@ matrix:
     - env: TOXENV=openapi
 
 services:
+  - docker
   - mongodb
+
+before_install:
+  - docker build -t xsnippet_api -f contrib/docker/Dockerfile .
+  - docker run -d --net=host xsnippet_api
 
 install:
   - pip install tox coveralls
 
 script:
   - tox
+  # TODO: replace the sanity check below with proper functional testing
+  - curl http://127.0.0.1:8000/snippets
 
 after_success:
   - coveralls
--- a/contrib/docker/Dockerfile
+++ b/contrib/docker/Dockerfile
@@ -1,0 +1,11 @@
+FROM python:latest
+MAINTAINER The XSnippet Team <dev@xsnippet.org>
+
+COPY . /app
+WORKDIR /app
+RUN pip install .
+
+ENV XSNIPPET_API_SETTINGS=/etc/xsnippet-api.conf
+
+EXPOSE 8000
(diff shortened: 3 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM python:3.5
+
+COPY . /app
+WORKDIR /app
+RUN pip install -e .
```
ikalnitsky: Why `-e`? I think if we're going to make a production ready solution, it'd be better to make an actual installation. What do you think?

Dockerfile
```diff
@@ -0,0 +1,10 @@
...
+
+ENV XSNIPPET_API_SETTINGS=/app/xsnippet_api/default.conf
```
ikalnitsky: If I'm not mistaken, `XSNIPPET_API_SETTINGS` should point to an additional external config. Default config will be used automatically, no need to point on it.

Dockerfile
malor: I don't have a preference here. Both must work.

Dockerfile
malor: Indeed. I just wanted to have an env variable for documentation purposes and looks like does not allow to leave it undefined.

ikalnitsky: Well, we can point it to some, let's say, `/etc/xsnippet-api.conf`. In that case, from API purpose of view all we need is to drop additional config into that location. What do you think?

Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM python:3.5
```
ikalnitsky: Please add `MAINTAINER` field below:

```
MAINTAINER The XSnippet Team <dev@xsnippet.org>
```

Currently `dev@xsnippet.org` forwards all incoming mails to my personal one, but I'm going to reconfigure it to send it to you too.

Dockerfile
```diff
@@ -0,0 +1,10 @@
...
+
+COPY . /app
+WORKDIR /app
+RUN pip install -e .
+
+ENV XSNIPPET_API_SETTINGS=/app/xsnippet_api/default.conf
+
+EXPOSE 8000
+CMD python -m xsnippet_api
```
ikalnitsky: It's recommended to use `ENTRYPOINT` instead of `CMD` for declaring default executable.

In case of `ENTRYPOINT`, we're able to pass any additional command line argument through `docker run` to `xsnippet_api`. Also, please consider to use `exec` form instead of `shell` one.

https://docs.docker.com/engine/reference/builder/#entrypoint

Dockerfile
malor: Done

Dockerfile
malor: Done

Dockerfile
malor: Done

PR #24, opened 2016-03-22
Add marker based pagination support to /snippets

malor:
It's now possible to pass additional ``limit`` and ``marker`` query
parameters to /snippets endpoint, e.g.:

    GET /snippets?limit=10&marker=1234

where ``limit`` will be an upper bound for the number of snippets to
be returned and ``marker`` is the ``id`` of the last snippet seen on
the previous page. Snippets are ordered by the creation order (newer
ones will be returned first).

Marker based pagination is used instead of limit/offset one, as the
performance of the latter degrades quickly as the number of entries
grows.

### DIFF
--- a/tests/resources/test_snippets.py
+++ b/tests/resources/test_snippets.py
@@ -32,8 +32,8 @@ class TestSnippets(metaclass=AIOTestMeta):
                 'tags': ['tag_a', 'tag_b'],
                 'author_id': None,
                 'is_public': True,
-                'created_at': now,
-                'updated_at': now,
+                'created_at': now - datetime.timedelta(100),
+                'updated_at': now - datetime.timedelta(100),
             },
             {
                 'id': 2,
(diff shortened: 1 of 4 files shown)


tests/resources/test_snippets.py
```diff
@@ -100,6 +100,82 @@ def _compare_snippets(self, snippet_db, snippet_api):
             for snippet_db, snippet_api in zip(expected, await resp.json()):
                 self._compare_snippets(snippet_db, snippet_api)
 
+    async def test_get_snippets_pagination(self):
+        now = datetime.datetime.utcnow().replace(microsecond=0)
+        snippet = {
+            'id': 3,
+            'title': 'snippet #3',
+            'content': '(println "Hello, World!")',
+            'syntax': 'clojure',
```
ikalnitsky: LOL :D

PR #27, opened 2016-09-06
Discover RESTful API via @endpoint decorator

ikalnitsky:
It's generally a more convenient to do API discovering rather than
importing resources one by one and adding it to router. That makes even
more sense in light of versioning API, where the very same resource may
appear in different routers.

This commit introduces an @endpoint decorator for resources. That allows
us to declaratively map a resource to some URI and defines an API
version range where it's applicable.

No tests are introduces, since those code is explicitly covered by any
test that use create_app() factory function.

### DIFF
--- a/xsnippet_api/application.py
+++ b/xsnippet_api/application.py
@@ -9,9 +9,12 @@
     :license: MIT, see LICENSE for details
 """
 
+import collections
+import pkg_resources
+
 from aiohttp import web, web_urldispatcher
 
-from . import database, resources, router
+from . import database, router
 
 
 def create_app(conf):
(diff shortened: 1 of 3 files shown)


xsnippet_api/application.py
```diff
@@ -27,19 +30,76 @@ def create_app(conf):
...
+        rv = collections.defaultdict(web_urldispatcher.UrlDispatcher)
+
+        # Create routers for each discovered API version. The main reason why
+        # we need that so early is to register resources in all needed routers
+        # according to supported version range.
+        for item in cls._registry:
+            rv[item.version]
+
+        for item in cls._registry:
+            # if there's no env_version then a resource is still working, and
```
malor: typo: end_version

xsnippet_api/application.py
```diff
@@ -27,19 +30,76 @@ def create_app(conf):
...
+            # latest discovered version should be considered as end_version
+            end_version = item.end_version or sorted(
+                rv.keys(),
+                key=lambda v: pkg_resources.parse_version(v)
```
malor: nit: can't we pass `pkg_resources.parse_version` directly? e.g.

```python

  key=pkg_resources.parse_version)

```

xsnippet_api/application.py
```diff
@@ -27,19 +30,76 @@ def create_app(conf):
...
+            )[-1]
+
+            supported_versions = [
+                version for version in rv.keys()
+                if pkg_resources.parse_version(item.version)
```
malor: nit: this could probably be more readable

xsnippet_api/application.py
ikalnitsky: Yep, you're right. At the very beginning I've been using another API, and lambda was required. :)

xsnippet_api/application.py
ikalnitsky: Any ideas how? I thought about

```
            V = pkg_resources.parse_version
            supported_versions = [
                version for version in rv.keys()
                if V(item.version) <= V(version) <= V(end_version)
            ]
```

So what do you think?

PR #28, opened 2016-09-12
POST /snippets - validate input data

ikalnitsky:
In order to prevent saving incorrect snippet in database we've got to
reject any requests with passed malformed snippet. We use Cerberus
library to check input data against Snippet schema.


xsnippet_api/resources/snippets.py
```diff
@@ -10,11 +10,206 @@
...
+import cerberus
 
 from .. import resource, services
 from ..application import endpoint
 
 
+# List of supported syntaxes. So far it's used only for validation purpose and
+# taking into account that syntax highlighting is going to be implemented on
+# client's side, we need to start consider removing that enum.
+_SYNTAXES = [
```
ikalnitsky: @malor What do you think about moving this list into `default.conf` ?

xsnippet_api/resources/snippets.py
```diff
@@ -10,11 +10,206 @@
...
+    'yaml',
+    'zephir',
+]
+
+_SCHEMA = {
+    'title': {'type': 'string'},
+    'content': {'type': 'string', 'required': True},
+    'syntax': {'type': 'string', 'allowed': _SYNTAXES},
+    'tags': {'type': 'list', 'schema': {'type': 'string', 'regex': '[\w_-]+'}},
+    'is_public': {'type': 'boolean', },
```
malor: nit: trailing comma in the literal is inconsistent with other values of this dict

xsnippet_api/resources/snippets.py
```diff
@@ -56,5 +251,16 @@ class Snippets(resource.Resource):
 
     async def post(self):
         snippet = await self.read_request()
+
+        v = cerberus.Validator(_SCHEMA)
```
ikalnitsky: I'm not sure about it. Take a look on how I retrieve errors below: I use `.error` attribute of validator, hence - it's mutable. I'm not sure it's cleared on each validation call, so I decided to create it from scratch each time.

Still, I need to investigate it.

PR #30, opened 2016-09-24
Add authentication middleware

malor:
The proposed middleware performs authentication of user requests by
validating tokens passed in X-Auth-Token header. The values are
expected to be in JSON Web Token format (RFC 7519).

The parsed token value (if validation succeeded) is then stored in
the request object to be used later in resource handlers.


tests/middlewares/test_auth.py
```diff
@@ -0,0 +1,77 @@
...
+    tests.middlewares.test_auth
+    ---------------------------
+
+    Tests XSnippet authentication middleware.
+
+    :copyright: (c) 2016 The XSnippet Team, see AUTHORS for details
+    :license: MIT, see LICENSE for details
+"""
+
+from aiohttp import web
```
ikalnitsky: nit:

```
import aiohttp.web as web
import jose.jwt as jwt
import pytest
import pkg_resources
```

just to have everything started from `import` :)

tests/middlewares/test_auth.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class TestAuthMiddleware(metaclass=AIOTestMeta):
+
+    conf = conf.get_conf(
+        pkg_resources.resource_filename('xsnippet_api', 'default.conf'))
+    conf['auth'] = {'secret': 'SWORDFISH'}
+
+    def setup(self):
+        self.app = application.create_app(self.conf)
```
ikalnitsky: I think it makes sense to create bare application instance since we ain't interested in XSnippet endpoints and so on. See how it's done in `tests/tests/test_resource.py`. In that case you won't even need to define and hijack custom resource.

tests/services/test_snippet.py
```diff
@@ -27,6 +27,8 @@ class TestSnippet(metaclass=AIOTestMeta):
     async def setup(self):
         conf = get_conf(
             pkg_resources.resource_filename('xsnippet_api', 'default.conf'))
+        conf['auth'] = {'secret': 'SWORDFISH'}
```
ikalnitsky: Hm.. Why do we need 'auth' on services layer? I thought it's required for API only. Could you please explain?

xsnippet_api/middlewares/auth.py
```diff
@@ -0,0 +1,47 @@
...
+    :copyright: (c) 2016 The XSnippet Team, see AUTHORS for details
+    :license: MIT, see LICENSE for details
+"""
+
+from aiohttp import web
+from jose import jwt
+
+
+async def auth(app, next_handler):
+    """Authentication middleware
```
ikalnitsky: nit: please add trailing dot.

xsnippet_api/application.py
```diff
@@ -58,7 +58,8 @@ def create_app(conf):
     # we need to import all the resources in order to evaluate @endpoint
     # decorator, so they can be collected and passed to VersionRouter
     from . import resources  # noqa
-    app = web.Application(router=router.VersionRouter(endpoint.collect()))
+    app = web.Application(router=router.VersionRouter(endpoint.collect()),
+                          middlewares=[middlewares.auth])
```
ikalnitsky: I wonder would it be possible to specify wanted middlewares in config file? Or once we're checking user's permissions we're tied to auth? Just can't figure out right now.

xsnippet_api/middlewares/auth.py
```diff
@@ -0,0 +1,47 @@
...
+        secret = app['conf']['auth']['secret']
+    except KeyError:
+        raise ValueError("Option 'secret' is required to be set when "
+                         "auth middleware is enabled")
+
+    async def auth_handler(request):
+        token = request.headers.get('X-Auth-Token')
+        if token is not None:
+            try:
+                request['auth'] = jwt.decode(token, secret)
```
ikalnitsky: Am I correct here that 

* `jwt.decode(...)` raises exception in case passed token is invalid.
* `jwt.decode(...)` returns human readable token including, for instance, user's permissions if it's valid, so we can check those in resources?

tests/middlewares/test_auth.py
malor: Done

xsnippet_api/middlewares/auth.py
malor: Done

xsnippet_api/application.py
malor: Yeah, I think it's one of those, that will always be enabled. The fact that it's a middleware helps keep code clean and modular, though.

tests/middlewares/test_auth.py
malor: Agreed. I still need additional resources, though, as they do asserts on the request object.

tests/services/test_snippet.py
malor: Ah, good catch: this one is not needed, indeed.

xsnippet_api/middlewares/auth.py
malor: That's correct. Subsequent changes will introduce the token format and service layer functions.

PR #34, opened 2017-01-22
Fix startup with default conf

malor:


xsnippet_api/middlewares/__init__.py
```diff
@@ -8,7 +8,7 @@
     :license: MIT, see LICENSE for details
 """
 
-from .auth import auth
+from . import auth
```
ikalnytskyi: Hm.. I think there's no value in preserving this importing stuff in case these stuff is modules. What do you say about making `__init__.py` empty?

xsnippet_api/middlewares/auth.py
```diff
@@ -9,10 +9,31 @@
...
 """
 
+import random
+import string
+
 import aiohttp.web as web
 import jose.jwt as jwt
 
 
+SECRET_LEN = 64
```
ikalnytskyi: Wow.. Security. :D

PR #37, opened 2017-04-03
Implement PUT & PATCH for /snippets/{id} endpoint

ikalnytskyi:
PUT & PATCH HTTP methods are intended to change snippet attributes by
authenticated owners. This is a first step towards changesets.

The difference between PUT and PATCH is in the following:

* PUT replaces snippet entity in the database. Any missed attributes will
  be reset to default.

* PATCH updates entity in the database. Any missed attributes will be
  kept untouched.

### DIFF
--- a/tests/resources/test_snippets.py
+++ b/tests/resources/test_snippets.py
@@ -384,10 +384,7 @@ async def test_post_snippet_malformed_snippet(name, value, testapp, snippets):
     assert resp.status == 400
 
     error_resp = await resp.json()
-    assert re.match(
-        'Cannot create a new snippet, passed data are incorrect. '
-        'Found issues: `%s` - .*' % name,
-        error_resp['message'])
+    assert re.match('`%s` - .*' % name, error_resp['message'])
 
 
 async def test_post_snippet_syntax_enum_allowed(
@@ -437,10 +434,7 @@ async def test_post_snippet_syntax_enum_not_allowed(
     assert resp.status == 400
 
     error_resp = await resp.json()
-    assert re.match(
-        'Cannot create a new snippet, passed data are incorrect. '
-        'Found issues: `syntax` - invalid value.',
-        error_resp['message'])
+    assert re.match('`syntax` - .*', error_resp['message'])
 
 
 async def test_data_model_indexes_exist(db):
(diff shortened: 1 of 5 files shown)


tests/resources/test_snippets.py
```diff
@@ -313,9 +330,7 @@ def _compare_snippets(self, snippet_db, snippet_api):
 
             error_resp = await resp.json()
             assert re.match(
-                'Cannot create a new snippet, passed data are incorrect. '
-                'Found issues: `syntax` - invalid value.',
-                error_resp['message'])
+                '`syntax` - unallowed value go.', error_resp['message'])
```
malor: lol :)

xsnippet_api/resources/snippets.py
```diff
@@ -16,44 +16,101 @@
...
+    })
+
+    if not v.validate(dict(resource.request.match_info)):
+        raise InvalidId('%s.' % cerberus_errors_to_str(v.errors))
+
+    return int(resource.request.match_info['id'])
+
+
+async def _write(resource, service_fn, *, status):
+    v = cerberus.Validator({
```
malor: I must have asked this before and it's probably not a big deal anyway, but can we create a Validator instance just once and reuse it rather than doing that on every call to _write()?

xsnippet_api/resources/snippets.py
```diff
@@ -16,44 +16,101 @@
...
-    'title': {'type': 'string'},
-    'content': {'type': 'string', 'required': True},
-    'syntax': {'type': 'string'},
-    'tags': {'type': 'list', 'schema': {'type': 'string', 'regex': '[\w_-]+'}},
-    'is_public': {'type': 'boolean'},
-    'author_id': {'type': 'integer', 'readonly': True},
-    'created_at': {'type': 'datetime', 'readonly': True},
-    'updated_at': {'type': 'datetime', 'readonly': True},
-}
+class InvalidId(Exception):
```
malor: Any specific reason not to put this into `xsnippet_api/exceptions.py`?

xsnippet_api/resources/snippets.py
```diff
@@ -16,44 +16,101 @@
...
+        return resource.make_response({'message': str(exc)}, status=404)
+
+    return resource.make_response(written, status=status)
+
+
+async def _read(resource, service_fn, *, status):
+    try:
+        read = await service_fn(_get_id(resource))
+
+    except InvalidId as exc:
```
malor: We should try to remove the code duplication, but it's not a blocker for now.

xsnippet_api/resources/snippets.py
```diff
@@ -16,44 +16,101 @@
...
         # TODO: only authorized owners can remove their snippets
-        return await self._run(services.Snippet(self.db).delete, 204)
+        return await _read(self, services.Snippet(self.db).delete, status=204)
 
-    async def _run(self, service_fn, status):
-        v = cerberus.Validator({
-            'id': {'type': 'integer', 'min': 1, 'coerce': try_int},
-        })
+    async def put(self):
+        async def service_fn(snippet):
```
malor: Not that it would change a whole lot, but `_write()` and `_read()` could be context managers, so what we would not need to build and pass a `service_fn`, but would just execute a code block within a context manager instead. Anyway, not critical either. We could revisit this later.

xsnippet_api/resources/snippets.py
ikalnytskyi: It's sort of local exception: it's used only there and outer world knows nothing about that. But maybe I overdesigned here and we should put it into exceptions.py.

xsnippet_api/resources/snippets.py
ikalnytskyi: Unfortunately, validator instance is mutable and set `.errors` attribute when validation is done. It looks like it resets this attribute between calls but it's not documented. I'm not sure it's a good idea to rely on this. What do you think?

xsnippet_api/resources/snippets.py
ikalnytskyi: I'm not sure how it will look like but I'll play around with that.

xsnippet_api/resources/snippets.py
malor: Makes sense. Let's keep it this way.

xsnippet_api/resources/snippets.py
malor: Ah, shit. Mutable things :( Let's keep it as is.

ikalnytskyi:
> We should also update the OpenAPI spec to include the new changes to the API.

I'm afraid OpenAPI spec is outdated already. I'm going to update it a little bit later in separate PR if you don't mind. 

> Maybe we should keep this disabled until we have a full story for auth?

Sure. I'll force this endpoints to require auth.

PR #40, opened 2017-04-19
Document API in OpenAPI specification

ikalnytskyi:
Current version of OpenAPI spec contains non-implemented parts of API as
well as missed some valuable information regarding authentication or
pagination. This commit removes unimplemented parts from the spec and
brings more details on other parts.

### DIFF
--- a/contrib/openapi/spec.yml
+++ b/contrib/openapi/spec.yml
@@ -1,453 +1,290 @@
 swagger: '2.0'
-info:
-  title: XSnippet API
-  description: REST API of XSnippet, a simple web-service for sharing code snippets
-  version: "4.0.0"
-
-host: api.xsnippet.org
-
 schemes:
   - http
-  - https
+host: api.xsnippet.org
+info:
+  title: XSnippet API
+  version: '1.0'
+  contact:
+    email: dev@xsnippet.org
+  description: |
+    The XSnippet API is organized around [REST]. Our API has predictable,
+    resource-oriented URLs, and uses HTTP response codes to indicate API
+    errors. We use built-in HTTP features, like HTTP authentication and
+    HTTP verbs, which are understood by off-the-shelf HTTP clients. See
+    more details below and feel free to contact us if any questions!
 
+    [REST]: https://en.wikipedia.org/wiki/Representational_state_transfer
+
+    Authentication
+    ==============
+
+    Authentication is done via [JSON Web Token] (JWT) that is issued by the
+    API in exchange for Google Access Token. The token needs to be passed
+    via `Authorization` HTTP header, and is mandatory for those endpoints
+    which require non-anonymous permissions.
+
+    ```http
+    Authorization: bearer {your-token-here}
+    ```
+
+    [JSON Web Token]: https://tools.ietf.org/html/rfc7519
+
+    Pagination
+    ==========
+
+    Pagination is done via two query parameters: `limit` and `marker`. The
+    first one limits a number of entries to be returned in response, while
+    the last one – defines your position in the list by taking an entry ID
+    and returning `limit` entries after that one assuming that entries are
+    sorted in some (defined by concrete API endpoint) order.
+
+    In other words if you make a list request and receive `100` entries,
+    ending with `foo`, your subsequent call can include `?marker=foo` in
+    order to fetch the next page of the list.
+
+    Versioning
+    ==========
+
+    API version is passed via non-standard `Api-Version` HTTP header, and
+    must be equal to `1.0` so far. In case the version is not passed, the
+    latest one will be assumed.
+
+    ```http
+    Api-Version: 1.0
+    ```
+  x-logo:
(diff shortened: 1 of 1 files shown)


contrib/openapi/spec.yml
```diff
@@ -1,453 +1,282 @@
...
   title: XSnippet API
-  description: REST API of XSnippet, a simple web-service for sharing code snippets
-  version: "4.0.0"
+  version: '1.0'
+  contact:
+    email: dev@xsnippet.org
+  description: |
+    The XSnippet API is organized around [REST]. Our API has predictable,
+    resource-oriented URLs, and uses HTTP response codes to indicate API
+    errors. We use built-in HTTP features, like HTTP authentication and
```
malor: HTTP authentication? I thought it was limited to standard methods (basic, digest etc)

contrib/openapi/spec.yml
```diff
@@ -1,453 +1,282 @@
...
+
+    In other words if you make a list request and receive `100` entries,
+    ending with `foo`, your subsequent call can include `?marker=foo` in
+    order to fetch the next page of the list.
+
+    Versioning
+    ==========
+
+    API version is passed via non-standard `Api-Version` HTTP header, and
+    must be equal to `1.0` so far. In case the version is not passed, the
```
malor: I wonder if we should default to the oldest version instead, e.g. to not surprise clients with a newly deployed update.

contrib/openapi/spec.yml
```diff
@@ -1,453 +1,282 @@
...
-      summary: Retrieve information about code snippets.
+      summary: List snippets
       description: |
-        The Snippets endpoint returns information about stored code snippets. The
-        listed snippets are sorted by the date-time of creation in the ascending order.
-
-        By default, only `limit` latest snippets are returned. Paginated queries
-        should be used in order to get more snippets.
+        Returned snippets are sorted in reversed chronological order, so
+        most recent ones appearing first.
```
malor: nit: s/appearing/appear

contrib/openapi/spec.yml
```diff
@@ -1,453 +1,282 @@
...
         format: date-time
-        description: Date-time value when the snippet was updated last time.
-
+        readOnly: true
+    required:
+      - content
   Error:
     type: object
     properties:
-      code:
```
malor: I think it's useful to have a "machine-readable" `code` value to specify error details besides the HTTP status, but as long as we don't have such examples in the implementation right now, we might add this back later.

contrib/openapi/spec.yml
```diff
@@ -1,453 +1,282 @@
...
+          description: A limit on the number of snippets to be returned.
       responses:
         200:
-          description: An array of code snippets
+          description: OK
           schema:
             type: array
             items:
               $ref: '#/definitions/Snippet'
-        400:
```
malor: `GET /snippets` can return `400 Bad Request` if `limit` / `marker` are incorrect. Why remove it here?

contrib/openapi/spec.yml
ikalnytskyi: According to [RFC-7235], HTTP provides a general framework for access control and authentication, via an extensible set of challenge-response authentication schemes, and `bearer` scheme is one of them (see [IANA Auth Schemes]). So I think it's correct to say that we support HTTP authentication. Otherwise, please provide better wording or suggest to remove it. :)

[RFC-7235]: https://tools.ietf.org/html/rfc7235
[IANA Auth Schemes]: https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml

contrib/openapi/spec.yml
ikalnytskyi: Thanks, I'll fix it.

contrib/openapi/spec.yml
ikalnytskyi: You are right, my mistake. I'll return both 400 and 404 (since we return 404 if `marker` points to a non-existing snippet).

contrib/openapi/spec.yml
ikalnytskyi: Agree. I just don't want to have something here as long as it's not implemented. If you think it should be here we can create a GitHub issue with proposal to add a machine-readable error code (so we don't forgot to add one).

contrib/openapi/spec.yml
malor: Ok, works for me, then :P Thanks for the links! ;)

contrib/openapi/spec.yml
ikalnytskyi: I'm not sure that defaulting to oldest version is a good idea either. You see if we default to oldest version then we are in the situation when we must support this version forever. Otherwise, we will break users anyway. Maybe the best option here is to force users to pass a version, but it will complicate API consuming via `curl` and company.

Nevertheless I just documented what we have right now. Could you please create a corresponding GitHub issue so we can move discussion and implementation there?

PR #43, opened 2017-10-22
Add support for filtering by title and tag

malor:
It's now possible to pass additional query parameters `title` and
`tag`, so that only the snippets that have corresponding attribute
values are returned, e.g.:

    GET /snippets?title=foo&tag=bar

Filtering by title is implemented by searching for all the matches,
that start with the given string (both for efficiency and
feature-parity reasons).

### DIFF
--- a/tests/services/test_snippet.py
+++ b/tests/services/test_snippet.py
@@ -121,6 +121,52 @@ class TestSnippet(metaclass=AIOTestMeta):
         excinfo.match(r'Sorry, cannot complete the request since `marker` '
                       r'points to a nonexistent snippet.')
 
+    async def test_get_filter_by_title(self):
+        result = await self.service.get(title='snippet #1')
+        assert result == [self.snippets[0]]
+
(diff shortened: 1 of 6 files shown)


tests/services/test_snippet.py
```diff
@@ -121,6 +121,49 @@ class TestSnippet(metaclass=AIOTestMeta):
...
+        assert prefix_match == list(reversed(self.snippets))
+
+        regexes_are_escaped = await self.service.get(title='^snippet.*')
+        assert regexes_are_escaped == []
+
+        nonexistent = await self.service.get(title='non existing snippet')
+        assert nonexistent == []
+
+    async def test_get_filter_by_tag(self):
+        result = await self.service.get(tag='tag_c')
```
ikalnytskyi: Maybe add also a test to ensure that the filtering works even for snippets with few tags?

malor: Done.

PR #45, opened 2017-10-22
Add a basic ansible playbook for deployment

malor:
The playbook allows to deploy xsnippet-api on a clean Ubuntu Xenial
machine, e.g.:

    $ ansible-playbook -u user \
                       -e xsnippet_image=xsnippet/xsnippet-api:latest \
        contrib/ansible/deploy.yaml

`xsnippet_image` variable can be used to specify the exact Docker
image, that should be deployed (it is expected, that the image has
already been built and pushed to the registry).

The playbook also deploys xsnippet-api dependencies like MongoDB.

### DIFF
--- a/contrib/ansible/deploy.yaml
+++ b/contrib/ansible/deploy.yaml
@@ -1,0 +1,68 @@
+---
+- hosts: xsnippet
+  tasks:
+    - name: ensure Python 2 is installed
+      raw: test -e /usr/bin/python || (apt -y update && apt install -y python-minimal)
+  gather_facts: false
+
+- hosts: xsnippet
+  vars:
+    xsnippet_image: xsnippet/xsnippet-api:latest
+  tasks:
+    - name: update apt cache (if needed)
+      apt: update_cache=yes cache_valid_time=3600
+
+    - name: install essential packages
+      apt: "name={{ item }} state=latest"
+      with_items:
+        - apt-transport-https
+        - ca-certificates
+
+    - name: add Docker gpg key
+      apt_key:
+        url: https://download.docker.com/linux/ubuntu/gpg
+        state: present
+
+    - name: add Docker packages repo
+      apt_repository:
+        repo: "deb [arch=amd64] https://download.docker.com/linux/ubuntu {{ ansible_distribution_release }} stable"
+        state: present
+      register: newrepo
+
+    - name: update apt cache (if needed)
+      apt: update_cache=yes
+      when: newrepo.changed
+
+    - name: install the latest Docker
+      apt: name=docker-ce state=latest
+
+    - name: initialize Docker Swarm
+      shell: "docker swarm init --advertise-addr 127.0.0.1 || docker stack ls"
+      register: init_result
+      changed_when: "'This node is already part of a swarm' not in init_result.stderr"
+
+    - name: create xsnippet user
+      user:
+        name: xsnippet
+        createhome: yes
+
+    - name: prepare the config file
+      template:
+        src: xsnippet-api.conf.j2
+        dest: "/home/xsnippet/xsnippet-api.conf"
+        mode: 0600
+        owner: xsnippet
+        group: xsnippet
+
+    - name: prepare the Docker stack config
+      template:
+        src: docker-compose.yml.j2
+        dest: "/home/xsnippet/docker-compose.yml"
+        mode: 0600
+        owner: xsnippet
+        group: xsnippet
+
+    - name: create or update a Docker Swarm stack
+      command: "docker stack deploy -c docker-compose.yml xsnippet"
+      args:
+        chdir: "/home/xsnippet"
--- a/contrib/ansible/docker-compose.yml.j2
+++ b/contrib/ansible/docker-compose.yml.j2
@@ -1,0 +1,38 @@
+version: "3.3"
+services:
+  db:
+    image: mongo:3
+    deploy:
+      replicas: 1
+      restart_policy:
+        condition: on-failure
+    volumes:
+      - db_state:/data/db
+    networks:
+      - default
+
+  api:
+    image: {{ xsnippet_image }}
+    configs:
+      - source: api_config
+        target: /etc/xsnippet-api.conf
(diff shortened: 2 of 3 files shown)


contrib/ansible/deploy.yaml
```diff
@@ -0,0 +1,62 @@
...
+          state: present
+
+    - name: update apt cache
+      apt: update_cache=yes
+
+    - name: install Docker
+      apt: name=docker-ce state=latest
+
+    - name: initialize Docker Swarm
+      shell: "docker swarm init --advertise-addr `ip -o a | grep eth | head -n1 | awk '{print $4}' | cut -d '/' -f 1` || docker stack ls"
```
ikalnytskyi: Jesus.. what does this line return?

```
`ip -o a | grep eth | head -n1 | awk '{print $4}' | cut -d '/' -f 1`
```

Can we use ansible facts instead to retrieve proper IP address?

contrib/ansible/deploy.yaml
```diff
@@ -0,0 +1,62 @@
...
+
+    - name: create xsnippet user
+      user:
+        name: xsnippet
+        groups: docker
```
ikalnytskyi: Hm.. I have one little concern about this group. Usually having this group means you have `sudo`-less permissions to use Docker. As we recently learned, this effectively means you have superuser permissions as you can create any binary inside container with `setuid 0`, and then call it on the system.

contrib/ansible/deploy.yaml
```diff
@@ -0,0 +1,62 @@
...
+        - ca-certificates
+        - curl
+        - software-properties-common
+
+    - name: add Docker gpg key
+      shell: "curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -"
+
+    - name: add Docker packages repo
+      apt_repository:
+          repo: "deb [arch=amd64] https://download.docker.com/linux/ubuntu {{ ansible_distribution_release }} stable"
```
ikalnytskyi: (minor) you use 2 space indentation everywhere in this playbook, so could you please fix these lines to have 2 spaces?

contrib/ansible/docker-compose.yml.j2
```diff
@@ -0,0 +1,38 @@
+version: "3.3"
+services:
+    db:
```
ikalnytskyi: (ever more minor) You use 2 space indentation in Ansible playbook, can you please use 2 space indention in this docker file? I think it's more than ok for YAMLs. :) But you can keep this way if you like, just a suggestion.

contrib/ansible/docker-compose.yml.j2
```diff
@@ -0,0 +1,38 @@
...
+            - db
+        ports:
+            - "80:8000"
+        networks:
+            - default
+networks:
+    default:
+
+volumes:
+    db_state:
```
ikalnytskyi: What does this empty object mean?

contrib/ansible/deploy.yaml
```diff
@@ -0,0 +1,62 @@
...
+  tasks:
+    - name: update apt cache
+      apt:  update_cache=yes cache_valid_time=3600
+
+    - name: install essential packages
+      apt: "name={{ item }} state=latest"
+      with_items:
+        - apt-transport-https
+        - ca-certificates
+        - curl
```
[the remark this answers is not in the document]
malor: No, that's it, but I'll replace it with ansible `apt_key` module as you're suggesting below.

contrib/ansible/deploy.yaml
```diff
@@ -0,0 +1,62 @@
...
+        - software-properties-common
```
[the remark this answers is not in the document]
malor: It provides the `apt-add-repository` util, but we don't really need it as we use an ansible module instead.

contrib/ansible/deploy.yaml
```diff
@@ -0,0 +1,62 @@
...
+
+    - name: add Docker gpg key
+      shell: "curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -"
```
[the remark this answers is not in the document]
malor: Thanks for suggestion! I didn't now it exists :)

contrib/ansible/deploy.yaml
malor: Done.

contrib/ansible/deploy.yaml
malor: Will do! (the first IPv4 address on the interfaces starting with `eth` :P )

contrib/ansible/deploy.yaml
malor: I just realized that we don't really need it. I'll remove it.

contrib/ansible/docker-compose.yml.j2
malor: Ok. Will do!

contrib/ansible/docker-compose.yml.j2
malor: Create a new volume named `db_state` with default parameters (default driver is `local`).

contrib/ansible/deploy.yaml
```diff
@@ -0,0 +1,62 @@
...
+  vars:
+    xsnippet_image: xsnippet/xsnippet-api:latest
+  tasks:
+    - name: update apt cache
+      apt:  update_cache=yes cache_valid_time=3600
+
+    - name: install essential packages
+      apt: "name={{ item }} state=latest"
+      with_items:
+        - apt-transport-https
```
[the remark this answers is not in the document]
zoresvit: Yeah, most distros still use `http` mirrors by default. Their reasoning behind this is that you don't need to trust a mirror, you only need to verify integrity of the packages (which is done relying on GPG).

While that's a valid justification, I still don't see a reason for them *not to use* `https` now when certificates are free and don't add to the cost of infrastructure. I know for Arch Linux it's possible to choose https-only mirrors just in case, not sure if the same holds for Ubuntu.

PR #47, opened 2017-12-21
Use xsnippet.api namespace instead of xsnippet_api

ikalnytskyi:
No particular reason. Just because @ikalnytskyi wants to! :P


ikalnytskyi:
> find_packages won't find sub-packages from a namespace because top-level dir doesn't have __init__.py, so they will have to be listed explicitly.

Wow, that's crap. They don't worth to be listed explicitly. :(

ikalnytskyi:
Geez, I can't reopen PR. :(

PR #52, opened 2017-12-26
Add GET /syntaxes endpoint

malor:
This new endpoint allows to retrieve the list of supported snippet
syntaxes, that is defined in the config file. Modify the default
config to contain the list of syntaxes from the original XSnippet.

Put the test fixtures into conftest.py, so that they can be reused
in tests of different API resources.

### DIFF
--- a/contrib/ansible/xsnippet-api.conf.j2
+++ b/contrib/ansible/xsnippet-api.conf.j2
@@ -35,7 +35,161 @@ connection = mongodb://db:27017/xsnippet
 ;        python
 ;        javascript
 ;
-; By default, constraint is not defined so any syntax is accepted.
+
+syntaxes =
+    APL
+    ASN.1
+    ASP.NET
+    Asterisk
+    Brainfuck
+    C
+    C#
+    C++
+    CMake
+    CQL
+    CSS
+    Clojure
+    ClojureScript
+    Closure Stylesheets (GSS)
+    Cobol
+    CoffeeScript
+    Common Lisp
+    Crystal
+    Cypher
+    Cython
+    D
+    DTD
+    Dart
+    Django
+    Dockerfile
+    Dylan
+    EBNF
+    ECL
+    Eiffel
+    Elm
+    Embedded Javascript
+    Embedded Ruby
+    Erlang
+    Esper
+    F#
+    FCL
+    Factor
+    Forth
+    Fortran
+    Gas
+    Gherkin
+    GitHub Flavored Markdown
+    Go
+    Groovy
+    HAML
+    HTML
+    HTTP
+    HXML
+    Haskell
+    Haskell (Literate)
+    Haxe
+    IDL
+    JSON
+    JSON-LD
+    JSX
+    Java
+    Java Server Pages
+    JavaScript
+    Jinja2
+    Julia
+    Kotlin
+    LESS
+    LaTeX
+    LiveScript
+    Lua
+    MS SQL
+    MUMPS
+    MariaDB SQL
+    Markdown
+    Mathematica
+    Modelica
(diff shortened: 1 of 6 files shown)


contrib/ansible/xsnippet-api.conf.j2
```diff
@@ -35,7 +35,102 @@ connection = mongodb://db:27017/xsnippet
 ;        python
 ;        javascript
 ;
-; By default, constraint is not defined so any syntax is accepted.
+
+syntaxes =
+    actionscript
```
ikalnytskyi: Is this a part of [codemirror syntaxes](https://github.com/codemirror/CodeMirror/tree/master/mode) ? :) Since we are going to use this library as both code editor and syntax highlighter, I'd suggest to use its mode as basis.

xsnippet/api/default.conf
```diff
@@ -35,7 +35,102 @@ connection = mongodb://localhost:27017/xsnippet
 ;        python
 ;        javascript
 ;
-; By default, constraint is not defined so any syntax is accepted.
+
+syntaxes =
```
ikalnytskyi: Do you have strong on opinion on filling `default.conf` with syntaxes? I'm just thinking that having it empty would be better for default as some might want to do not restrict submitted snippets to this set of syntaxes. Even more, we might want to change the highlighting library on front-end side, hence the list of syntaxes will change accordingly.

xsnippet/api/resources/syntaxes.py
```diff
@@ -0,0 +1,23 @@
...
+    :copyright: (c) 2017 The XSnippet Team, see AUTHORS for details
+    :license: MIT, see LICENSE for details
+"""
+
+from .. import resource
+from ..application import endpoint
+
+
+@endpoint('/syntaxes', '1.0')
+class Syntax(resource.Resource):
```
ikalnytskyi: Maybe call it `Syntaxes`? the same way as the endpoint?

tests/resources/test_syntaxes.py
```diff
@@ -0,0 +1,62 @@
...
+
+    Tests Syntaxes resource.
+
+    :copyright: (c) 2017 The XSnippet Team, see AUTHORS for details
+    :license: MIT, see LICENSE for details
+"""
+
+import pytest
+
+
```
ikalnytskyi: Can we also ensure expected behavior when no `syntaxes` set in config? Maybe respond with empty array or so?

contrib/ansible/xsnippet-api.conf.j2
malor: Ok, will do!

tests/resources/test_syntaxes.py
malor: Ah, good catch! This case wasn't covered...

xsnippet/api/default.conf
malor: No, I just thought it would make the migration easier. We can perform when it's time.

xsnippet/api/resources/syntaxes.py
malor: Will do!

ikalnytskyi:
Good job!

contrib/ansible/xsnippet-api.conf.j2
```diff
@@ -35,7 +35,129 @@ connection = mongodb://db:27017/xsnippet
...
+    powershell
+    properties
+    protobuf
+    pug
+    puppet
+    python
+    q
+    r
+    rpm
+    rpm/changes
```
ikalnytskyi: Seems like an artifact. :)

PR #53, opened 2018-01-03
Remove router assignment via app._router property

ikalnytskyi:
We use to set VersionRouter instance on application later in the code by
exploiting private attribute. The need of this hack comes from the time
when UrlDispatcher (router) was needed to be aware of the application,
and the application was needed to be aware of router on creation time.

In aiohttp v2.3.5 (we depend on) it's no more the case. So let's remove
the hack and use intended design.


ikalnytskyi:
Honestly, I didn't intended to do it this way.. I made it just to make it easier to review the code - make an PR into this branch allowed me to do not show commits from this branch on review. :)

PR #58, opened 2018-01-08
Provide Link response header for pagination

malor:
When traversing the list of snippets page by page provide an additional
response header - Link - that contains the links to the previous and
the next page with the given value of limit (if such pages exist).

This follows the idea from:

    https://developer.github.com/v3/guides/traversing-with-pagination/

and

    https://tools.ietf.org/html/rfc5988

Hopefully, this will make using of pagination easier on the client
side.

### DIFF
--- a/tests/resources/test_snippets.py
+++ b/tests/resources/test_snippets.py
@@ -267,7 +267,7 @@ async def test_get_snippets_filter_by_title_and_tag_with_pagination(
         _compare_snippets(snippet_db, snippet_api)
     # Check that urls in Link preserve the additional query params
     expected_link1 = (
-        '<https://api.xsnippet.org/snippets?title=snippet+%231&tag=tag_a&limit=1>; rel="first", '  # noqa
+        '<https://api.xsnippet.org/snippets?title=snippet+%231&tag=tag_a&limit=1>; rel="first", '
         '<https://api.xsnippet.org/snippets?title=snippet+%231&tag=tag_a&limit=1&marker=3>; rel="next"'  # noqa
     )
     assert resp.headers['Link'] == expected_link1
@@ -285,8 +285,8 @@ async def test_get_snippets_filter_by_title_and_tag_with_pagination(
         _compare_snippets(snippet_db, snippet_api)
     # Check that urls in Link preserve the additional query params
     expected_link2 = (
-        '<https://api.xsnippet.org/snippets?title=snippet+%231&tag=tag_a&limit=1>; rel="first", '  # noqa
-        '<https://api.xsnippet.org/snippets?title=snippet+%231&tag=tag_a&limit=1>; rel="prev"'  # noqa
+        '<https://api.xsnippet.org/snippets?title=snippet+%231&tag=tag_a&limit=1>; rel="first", '
+        '<https://api.xsnippet.org/snippets?title=snippet+%231&tag=tag_a&limit=1>; rel="prev"'
     )
     assert resp.headers['Link'] == expected_link2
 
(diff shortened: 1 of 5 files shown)


tests/resources/test_snippets.py
```diff
@@ -320,6 +336,244 @@ def _compare_snippets(snippet_db, snippet_api):
         _compare_snippets(snippet_db, snippet_api)
 
 
+# a helper function of traversing of the list of snippets via API
+async def get_next_page(testapp, limit=3, marker=0):
```
ikalnytskyi: (minor) If it's a helper function, maybe prefix it with an underscore?

tests/resources/test_snippets.py
```diff
@@ -258,21 +258,37 @@ def _compare_snippets(snippet_db, snippet_api):
...
         headers={
             'Accept': 'application/json',
+            'Host': 'api.xsnippet.org',
+            'X-Forwarded-Proto': 'https',
         })
     assert resp.status == 200
     expected = [snippets[0]]
     for snippet_db, snippet_api in zip(expected, await resp.json()):
         _compare_snippets(snippet_db, snippet_api)
+    # check that urls in Link preserve the additional query params
```
ikalnytskyi: (minor) I used capitalize sentences in comments around the code. It'd be nice to follow this style if you don't mind. :) So you can ignore this one really, just FYI.

tests/resources/test_snippets.py
```diff
@@ -258,21 +258,37 @@ def _compare_snippets(snippet_db, snippet_api):
...
+    expected_link2 = (
+        '<https://api.xsnippet.org/snippets?title=snippet+%231&tag=tag_a&limit=1>; rel="first", '  # noqa
```
ikalnytskyi: If you prefer, we can increase max-line-length up to 100 characters. Actually, PEP8 says it's ok to have lines up to 99 characters sometimes, it's just should not be a general rule.

tests/resources/test_snippets.py
```diff
@@ -320,6 +336,244 @@ def _compare_snippets(snippet_db, snippet_api):
...
+
+    # we should have seen snippets with ids 10, 9 and 8. No link to the prev
+    # page, as we are at the very beginning of the list
+    resp1 = await get_next_page(testapp, limit=3)
+    expected_link1 = (
+        '<https://api.xsnippet.org/snippets?limit=3>; rel="first", '
+        '<https://api.xsnippet.org/snippets?limit=3&marker=8>; rel="next"'
+    )
+    assert resp1.headers['Link'] == expected_link1
+    assert [s['id'] for s in await resp1.json()] == [10, 9, 8]
```
ikalnytskyi: 👍

tests/resources/test_snippets.py
```diff
@@ -320,6 +336,244 @@ def _compare_snippets(snippet_db, snippet_api):
...
+    resp3 = await get_next_page(testapp, limit=4, marker=5)
+    expected_link3 = (
+        '<https://api.xsnippet.org/snippets?limit=4>; rel="first", '
+        '<https://api.xsnippet.org/snippets?limit=4&marker=9>; rel="prev"'
+    )
+    assert resp3.headers['Link'] == expected_link3
+    assert [s['id'] for s in await resp3.json()] == [4, 3, 2, 1]
+
+
+async def test_pagination_links_non_consecutive_ids(testapp, db):
```
ikalnytskyi: Great! That's great!

tests/resources/test_snippets.py
malor: Done!

tests/resources/test_snippets.py
malor: Done!

tests/resources/test_snippets.py
malor: Done!

PR #60, opened 2018-01-10
Tweak the default acccess log format

### LINKED ISSUE #59: Service logs refer to reverse proxy IP instead of a client's one
Reading service logs on production with `docker service logs -f xsnippet_api` gives you:

```
xsnippet_api.2.rer8c21yct4f@xsnippet    | INFO:aiohttp.access:10.0.0.13 - - [10/Jan/2018:06:37:42 +0000] "GET /snippets?title=clojure HTTP/1.0" 200 267 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/63.0.3239.132 Safari/537.36"
```

Note, that `10.0.0.13` is the IP address of a Nginx running in a Docker container in front of the API service, and the IP address of the remote client, that we are interested in.

malor:
Make sure we've got the most important information like remote addr,
request status line, response code, response size, user agent and
request processing time, e.g:

```
[10/Jan/2018:21:00:14 +0000] 127.0.0.1 "GET /snippets?limit=5 HTTP/1.1" 200 233 curl/7.54.0" 0.030207
```

Note, that for production deploys behind a reverse-proxy this format
might be changed to use the HTTP headers like X-Real-IP or
X-Forwarded-For, if needed.

Partially closes #59

### DIFF
--- a/xsnippet/api/default.conf
+++ b/xsnippet/api/default.conf
@@ -15,6 +15,22 @@ host = 127.0.0.1
 port = 8000
 
 
+; ACCESS LOG FORMAT
+;
+; Note, that you have to use double % signs for escaping to work.
+;
+; %t  - datetime of the request
+; %a  - remote address
+; %r  - request status line
+; %s  - response status code
+; %b  - response size (bytes)
+; %Tf - request time (seconds)
+;
+; When deployed behind a reverse proxy, consider using the value of
+; headers like X-Real-IP or X-Forwarded-For instead of %a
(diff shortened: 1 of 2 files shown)


xsnippet/api/default.conf
```diff
@@ -15,6 +15,22 @@ host = 127.0.0.1
 port = 8000
 
 
+; ACCESS LOG FORMAT
```
ikalnytskyi: Oh gosh, we've got to move away from conf and use env variables as soon as possible, or it will be too late. :)

PR #61, opened 2018-01-13
Make Resource class more "RESTful"

ikalnytskyi:
RESTfulness is weird word, I know, but the idea is to provide a base
class for purposes of building RESTful APIs. Gosh, it sounds complex
so here's what was changed:

 * Resource classes now return tuples instead of 'web.Response()' of
   the following format: (data, status_code, headers). Where both
   status_code and headers are optional and can be omitted. From now
   one, content negotiation is done by resource class implicitly.

 * Exception translation to proper HTTP response is done by Resource
   class either. So far domain exceptions are hardcoded now, but they
   will become optional very soon once I figure out the better API for
   parametrization. :)

### DIFF
--- a/xsnippet/api/resource.py
+++ b/xsnippet/api/resource.py
@@ -10,6 +10,7 @@
     :license: MIT, see LICENSE for details
 """
 
+import asyncio
 import json
 import datetime
 import fnmatch
@@ -17,6 +18,8 @@ import functools
 
 from aiohttp import web, hdrs
 from werkzeug import http
+
+from . import exceptions
 
 
 class _JSONEncoder(json.JSONEncoder):
@@ -35,7 +38,7 @@ class _JSONEncoder(json.JSONEncoder):
 
 
 class Resource(web.View):
-    """Resource wrapper around :class:`aiohttp.web.View`.
+    """Class-based view to represent RESTful resource.
 
     The class provides basic facilities for building RESTful API, such as
     encoding responses according to ``Accept`` and decoding requests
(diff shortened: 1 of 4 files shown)


xsnippet/api/resource.py
```diff
@@ -58,7 +61,37 @@ def __init__(self, *args, **kwargs):
...
 
-    def make_response(self, data, headers=None, status=200):
+    @asyncio.coroutine
+    def __iter__(self):
+        # So far (Jan 5, 2018) aiohttp doesn't support custom request classes,
+        # but we would like to provide a better user experience for consumers
+        # of this middleware. Hence, we monkey patch the instance and add a new
+        # async method that would de-serialize income payload according to HTTP
+        # content negotiation rules.
+        setattr(self.request.__class__, 'get_data', self._get_data())
```
malor: I know that some people would not approve, but I do not think we should follow all OOP rules :P This seems to be a simpler solution.

xsnippet/api/resource.py
```diff
@@ -58,7 +61,37 @@ def __init__(self, *args, **kwargs):
...
+
+        try:
+            response = yield from super(Resource, self).__iter__()
```
malor: Yay for the beauty of asyncio!

xsnippet/api/resource.py
ikalnytskyi: It's just for start. I think we can come up with proxy object once it's a standalone solution (in separate package) but for the sake of simplicity now I decided to stick with monkey-patching. :)

PR #63, opened 2018-01-17
Implement changesets on database level

ikalnytskyi:
Changesets are upcoming feature of XSnippet that allows to track changes
to snippets made by their owners. In order to avoid number of database
migrations, let's move onto new scheme right now and introduce basic
support.

The code is ugly now, yes, but I promise to refactor it in near future!

### DIFF
--- a/tests/resources/test_snippets.py
+++ b/tests/resources/test_snippets.py
@@ -18,13 +18,15 @@ import pytest
 
 
 @pytest.fixture(scope='function')
-def snippets():
+def snippets(db):
     now = datetime.datetime.utcnow().replace(microsecond=0)
     return [
         {
             'id': 1,
             'title': 'snippet #1',
-            'content': 'def foo(): pass',
+            'changesets': [
+                {'content': 'def foo(): pass'},
+            ],
             'syntax': 'python',
             'tags': ['tag_a', 'tag_b'],
             'created_at': now - datetime.timedelta(100),
@@ -33,7 +35,9 @@ def snippets():
         {
             'id': 2,
             'title': 'snippet #2',
-            'content': 'int do_something() {}',
+            'changesets': [
+                {'content': 'int do_something() {}'},
+            ],
             'syntax': 'cpp',
             'tags': ['tag_c'],
             'created_at': now,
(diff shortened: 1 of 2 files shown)


xsnippet/api/services/snippet.py
```diff
@@ -47,23 +47,41 @@ def __init__(self, db):
...
         snippet['id'] = snippet_id
+        snippet['content'] = snippet.pop('changesets', [])[0]['content']
+
         return snippet
 
     async def update(self, snippet):
         now = datetime.datetime.utcnow().replace(microsecond=0)
         snippet['updated_at'] = now
 
+        condition = {'$set': snippet}
```
malor: nit: it's not a condition, it's the `update` parameter - a document that describes what updates must be performed

xsnippet/api/services/snippet.py
```diff
@@ -111,7 +129,11 @@ def __init__(self, db):
             ('created_at', sort['created_at']),
             ('_id', sort['_id']),
         ])
-        return await query.limit(limit).to_list(None)
+
+        snippets = await query.limit(limit).to_list(None)
+        for snippet in snippets:
+            snippet['content'] = snippet.pop('changesets', [])[-1]['content']
```
malor: nit: maybe we should leave this for the API layer, but we can discuss this later, when we actually start using the changesets

tests/resources/test_snippets.py
```diff
@@ -18,13 +18,15 @@
 
 
 @pytest.fixture(scope='function')
-def snippets():
+async def snippets(db):
```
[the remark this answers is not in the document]
ikalnytskyi: It happens due to `pytest-aiohttp` that provides necessary hooks to run this :)

xsnippet/api/services/snippet.py
ikalnytskyi: Will rename

xsnippet/api/services/snippet.py
ikalnytskyi: I was thinking about it and agree it probably would be better this way.. I don't remember why I decided to put it here for now, but I bet there were a reason. I think we will change this place plenty of times while we really start using it :)

Issue #41, opened 2017-04-23
Resolve the ambiguity of the missing `Api-Version` header


malor:
If the `Api-Version` header value is not passed explicitly, there are multiple options how the  API request could be handled:

1.  Forbid such HTTP requests and *require* the `Api-Version` header to be passed.

Pros:
- no ambiguity in the first place

Cons:
- requires passing a version value explicitly, even if clients do not really care (e.g. minor changes between versions or the version in question did not change this particular API endpoint)
- breaks existing API clients (if there are any :))

2. Treat the missing header value as if the *minimum* supported version was specified. 

Pros:
- deployment of new versions to production does not break the clients that do not specify the version value explicitly for each request

Cons:
- clients still break when the minimum supported API version is bumped on the server side (it's hard to come up with a *visible* deprecation mechanism - warnings in responses can easily be ignored as well as deprecation markers in the version list)

3. Treat the missing header value as if the *maximum* supported version was specified.

Pros:
- effectively forces the clients to follow the updates of the server code

Cons:
- every new deployment to production potentially breaks the existing clients that do not specify the version value explicitly for each request

We should choose the way how we want this to be solved and update the code / the OpenAPI spec accordingly.

zoresvit:
Move fast and break things! © 👍 😃 

I vote for option 3.

Rationale
---------

I think every interface should aim to provide maximum functionality with minimum user effort (learning curve). So the less knowledge user needs to use the interface — the better. It also needs to be convenient and safe.

With option 1 the user has to have prior knowledge about required headers (steeper learning curve) and has to specify the header with every call (inconvenient, especially with `curl`/`httpie`). In such scenario the very first attempt of the user to use the API intuitively (`curl api.xsnippet.org`) is ending with an error instructing to specify the header. It's also guaranteed to be a breaking change for existing clients. Since it's inconvenient and breaking, seems like a lose-lose to me. This already excludes option 1.

We are left with options 2 and 3. Option 2 might seem as a viable choice to make the API safe, but if we look at it carefully — it's just a delayed version of option 3. Both 2 and 3 _might_ eventually break existing client, option 2 will just do it a bit later, so for *existing* users there's no much difference. However for new users option 2 has major disadvantage: by default it sends *new* users to an old API version. This deprives new users of benefits of new API version (new features, fixes, performance improvements), while doesn't provide much more stability to existing users (if there is a breaking change, it will still break clients, just later). So essentially option 2 just makes users to live in the past :)

[...]

PR #64, opened 2018-01-20
Use explicitly set API version when not passed

ikalnytskyi:
We used to fallback to latest stable API version when 'Api-Version'
header was omitted. While it made sense when we were using version
discovery mechanism, it makes a little sense now when all versions
are statically defined.

Since now it's more garbage and complication, not saying it doesn't
give us control on which version to serve by default, let's remove
this function now.

### DIFF
--- a/tests/test_router.py
+++ b/tests/test_router.py
@@ -14,95 +14,63 @@ from aiohttp import web
 from xsnippet.api import router
 
 
-class TestVersionRouter:
+@pytest.fixture(scope='function')
+async def testapp(test_client):
+    class _TestResource1(web.View):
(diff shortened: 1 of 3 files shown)


malor:
Oh, and #41 can be closed now.

ikalnytskyi:
@malor Actually I'm not sure it's closed, as #41 is about convention to follow (which api version to choose as default) while this is an intermediate step to make things simpler now. :)

malor:
Yeah, I guess I meant that now we have to explicitly choose the version to be used in the code. So it's not like we've added a new one and it automatically has become the "latest".

PR #66, opened 2018-01-21
Use DI to inject database/conf where needed

ikalnytskyi:
So far database connection available as resource's 'db' attribute and we
passed it down to service layer.  Similarly, we extract 'conf' out from
application which is extracted out of request. It works, and it works
good, however code readability is not so good as we would like.

In order to improve code readability and provide some sort of connection
points and ask for a 'database' and 'conf' only where needed, let's try
to use dependency injection framework to provide those.

Unfortunately, some barely related changes are included in this PR, so
may Roman forgive me this. :)


xsnippet/api/database.py
```diff
@@ -43,17 +43,13 @@ def create_connection(conf):
     return db
 
 
-async def setup(app):
+async def setup(app, db):
     """Perform setup actions on application startup.
 
     This function is expected to inject the `db` key to the application
```
malor: nit: this is no longer true

xsnippet/api/middlewares/auth.py
```diff
@@ -19,19 +19,17 @@
 SECRET_LEN = 64
 
 
-async def setup(app):
+async def setup(app, conf):
```
malor: nit: we no longer need the `app` argument

xsnippet/api/application.py
```diff
@@ -65,12 +70,9 @@ def create_app(conf):
...
         ],
         router=router.VersionRouter({'1.0': v1}, default='1.0'))
-    app.on_startup.append(middlewares.auth.setup)
 
-    # Attach settings to the application instance in order to make them
-    # accessible at any point of execution (e.g. request handling).
-    app['conf'] = conf
-    app.on_startup.append(database.setup)
+    app.on_startup.append(functools.partial(middlewares.auth.setup, conf=conf))
+    app.on_startup.append(functools.partial(database.setup, db=db))
```
malor: I wonder if the app should be concerned with the database connection at all, now, that we do not need to carry a reference to a database connection around, but can just consume it from picobox. On the other hand,  we do need to call this `setup()` function from somewhere, so maybe it's just ok.

ikalnytskyi: Yeah, I have the very same feeling about this. Honestly, I just don't know yet where it would be better to put, so I decided to go with this for now and think about it a little bit later.

One thing I know for sure, we do not need `middlewares.auth.setup` to be called from here. It generates a secret when it's not set, so I'm thinking we should move it away from `create_app` and maybe put in `__main__`. Besides, doing so will allow us to have `auth` module untied from the app, hence make it reusable.

xsnippet/api/middlewares/auth.py
ikalnytskyi: That's right, however, it's passed by `aiohttp` anyway, so we can't just ignore it. :(

xsnippet/api/database.py
ikalnytskyi: Yeah, I'll fix it.

PR #67, opened 2018-01-21
Use werkzeug.MIMEAccept() to find best MIME match

ikalnytskyi:
According to HTTP standard, 'Accept' header may enlist multiple media
ranges, each may come with a priority qualifier. That means we need to
do proper parsing and picking the one that matches best.

The general rules are:

 * Media ranges without priority qualifier assume to have priority q=1.
 * Media ranges with globes (e.g. text/*) assume to have lower priority
   than media ranges without globes if the former have priority q=1.
 * Priority is a floating number in [0; 1] range, where 1 is higher
   priority and 0 is lower.

To do such a complex parsing, we use werkzeug's parse_accept_header that
returns all media ranges sorted by priority and then simply match them
one by one.

However, some simplification can be achieved by using werkzeug's
MIMEAccept class that does matching inside and returns the best match
value. This commit replaces custom matching code with werkzeug's
convenience and also adds a bunch of tests to check this functionality
we previously missed.

### DIFF
--- a/tests/test_resource.py
+++ b/tests/test_resource.py
@@ -9,6 +9,7 @@
 """
 
 import json
+import collections
 
 import aiohttp.web as web
 import pytest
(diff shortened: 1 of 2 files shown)


tests/test_resource.py
```diff
@@ -29,24 +30,29 @@ class _TestResource(resource.Resource):
 @pytest.fixture(scope='function')
 async def testapp(test_client):
     app = web.Application()
-    app['db'] = None
```
malor: :)

tests/test_resource.py
```diff
@@ -55,6 +61,49 @@ class _TestResource(resource.Resource):
         assert resp.status == 406
 
 
+@pytest.mark.parametrize('accept, best_match', [
```
malor: \o/

PR #70, opened 2018-01-22
Move Dockerfile to the root to simplify integration with Dockerhub

malor:
Automated builds on Dockerhub do not allow to specify the build
context location and would prefer to have Dockerfile in the root
of the repository working directory checkout.

If we do not have any strong preferences, it would be easier if
we just went that way.


.travis.yml
```diff
@@ -18,7 +18,7 @@ services:
   - mongodb
 
 before_install:
-  - docker build -t xsnippet_api -f contrib/docker/Dockerfile .
+  - docker build -t xsnippet_api -f Dockerfile .
```
ikalnytskyi: I don't think there's much sense to explicitly specify `-f Dockerfile` 'cause it's default. Though, doesn't matter much. :)

ikalnytskyi: Just a thought: what if we would use `tox -e docker-build` here?

PR #71, opened 2018-01-22
Use autogenerated IDs in tests

ikalnytskyi:
Due to custom SON manupulator we use that implements ID autoincrement,
it's not safe to manually set IDs on the documents because insertion of
such documents won't update internal counter. That basically means that
any further attempt to insert without ID (autoincrement case) may fail
with collision error (duplicated ID).

Since we don't actually care about pre-hardcoded IDs in the tests, let's
move to autogenerated IDs for the sake of simplicity.

### DIFF
--- a/tests/resources/test_snippets.py
+++ b/tests/resources/test_snippets.py
@@ -22,7 +22,6 @@ async def snippets(testdatabase):
     now = datetime.datetime.utcnow().replace(microsecond=0)
     snippets = [
         {
-            'id': 1,
             'title': 'snippet #1',
             'changesets': [
                 {'content': 'def foo(): pass'},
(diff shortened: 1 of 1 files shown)


tests/resources/test_snippets.py
```diff
@@ -44,7 +42,19 @@
             'updated_at': now,
         },
     ]
-    await testdatabase.snippets.insert(snippets)
+
+    # Usually pymongo updates passed collection so it has generated IDs
+    # in-place. However, due to our custom SON manipulators it's not the
+    # case in our case, since one of them makes a shallow copy which
```
malor: "it's not the case in our case" sounds funny :)

PR #73, opened 2018-01-30
Refactor test_snippets.py to be straightforward

ikalnytskyi:
Gosh, it's hard to tell what was wrong except for internal feelings.
There was annoying _compare_snippets() methods and database calls in
methods. There was annoying pop() calls to snippet at the beginning of
tests. There were other bad smell things.

The main idea behind this refactoring is to increase test readability
and avoid test duplication by using test parametrization.

### DIFF
--- a/tests/resources/test_snippets.py
+++ b/tests/resources/test_snippets.py
@@ -9,17 +9,60 @@
 """
 
 import re
-import copy
 import json
 import datetime
-import operator
+import cgi
+import urllib.parse
 
 import pytest
 
 
+class _pytest_regex:
+    """Assert that a given string matches a given regular expression."""
+
+    def __init__(self, pattern, flags=0):
+        self._regex = re.compile(pattern, flags)
+
+    def __eq__(self, actual):
+        return bool(self._regex.match(actual))
+
+    def __repr__(self):
+        return self._regex.pattern
+
+
+class _pytest_link_header:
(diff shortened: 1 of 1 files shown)


tests/resources/test_snippets.py
```diff
@@ -58,288 +67,177 @@
...
-        '/snippets?tag=' + value,
-        headers={
-            'Accept': 'application/json',
-        })
-    assert resp.status == 400
-    assert await resp.json() == {
-        'message': "`tag` - value does not match regex '[\\w_-]+'."
-    }
-
+@pytest.mark.parametrize('params, rv, link', [
```
malor: (here and below) maybe we should use the `ids` param to give these cases explicit names?

tests/resources/test_snippets.py
```diff
@@ -9,17 +9,60 @@
...
+
+            rv.append((parts.scheme, parts.netloc, parts.path, query, params))
+
+        # Dictionaries can't be compared by < and/or >, so we need to sort by
+        # first three components (scheme, netloc, path) and optional "rel"
+        # parameter because the latter usually uniquely defines a link.
+        return sorted(rv, key=lambda i: (i[-1].get('rel'), i[:3]))
+
+    def __init__(self, link):
+        self._link = link
```
malor: nit: you could parse the given link once and save the result here

PR #76, opened 2018-03-11
Generate secret in main rather than in middleware

ikalnytskyi:
There are couple of reasons of doing so but the main one is to make
middleware to be responsible for validating tokens only, moving
responsibility of getting secret (reading from config or generating one)
from middleware (as it's supposed to be as general as possible) to
consumers.

It'd make even more sense once we implement /auth endpoint that would be
responsible for issuing JWT tokens, meaning that the middleware is not
the only owner of authentication.

### DIFF
--- a/xsnippet/api/__main__.py
+++ b/xsnippet/api/__main__.py
@@ -9,9 +9,11 @@
     :license: MIT, see LICENSE for details
 """
 
+import binascii
 import logging
 import os
 import sys
+import warnings
 
 import aiohttp.web as web
 import picobox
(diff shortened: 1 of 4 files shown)


ikalnytskyi:
@malor I wonder what do you think about tests? Should I add some `test_main.py`, mock everything around and check that the secret is generated? Or it's something we can omit and ensure it's implicitly covered, let's say, by functional tests (i.e. we run in docker server and make curl request in .travis.yml) ?

xsnippet/api/__main__.py
```diff
@@ -31,6 +33,16 @@ def main(args=sys.argv[1:]):
...
     database = create_connection(conf)
 
+    # The secret is required to sign issued JWT tokens, therefore it's crucial
+    # to warn about importance of settings secret before going production. The
+    # only reason why we don't enforce it's setting is because we want the app
+    # to fly (at least for development purpose) using defaults.
+    if not conf['auth'].get('secret', ''):
+        warnings.warn(
+            'Auth secret has not been provided. Please generate a long and '
+            'random secret before going production.')
```
malor: typo: to production

xsnippet/api/__main__.py
```diff
@@ -31,6 +33,16 @@ def main(args=sys.argv[1:]):
     ], envvar='XSNIPPET_API_SETTINGS')
     database = create_connection(conf)
 
+    # The secret is required to sign issued JWT tokens, therefore it's crucial
+    # to warn about importance of settings secret before going production. The
+    # only reason why we don't enforce it's setting is because we want the app
+    # to fly (at least for development purpose) using defaults.
+    if not conf['auth'].get('secret', ''):
+        warnings.warn(
+            'Auth secret has not been provided. Please generate a long and '
```
malor: nit: a long random

(w/o "and")

malor:
> @malor I wonder what do you think about tests? Should I add some test_main.py, mock everything around and check that the secret is generated? Or it's something we can omit and ensure it's implicitly covered, let's say, by functional tests (i.e. we run in docker server and make curl request in .travis.yml) ?

I'd go for functional tests.

PR #77, opened 2018-03-11
Do not default 'syntax' to 'text' if not passed

ikalnytskyi:
We used to have sort of weird behaviour by defaulting snippet's 'syntax'
to 'text' if it wasn't specified at the first place. While it'd make
sense on certain situations, it's something we want to avoid in
XSnippet, because we might want to know whether it's something submitted
by user or not.

Clearly there are two options to go with:

 (a) Require 'syntax' to be explicitly passed in POST and PUT requests.
 (b) Assume 'syntax' is 'None' if not passed.

In case of option (a) we have an open question: what to do if we didn't
set a list of allowed syntaxes? Should we force users to pass any syntax
even if it doesn't matter?

In case of option (b) an open question is opposite: what to do if we set
a list of allowed syntaxes? Should we ignore this implicit 'syntax'
value and allow to store 'None'?

Despite not having a good way to go, we must come up with some solution.
So this commit implements option (b).

### DIFF
--- a/tests/resources/test_snippets.py
+++ b/tests/resources/test_snippets.py
@@ -521,26 +521,35 @@ async def test_get_snippets_pagination_not_found(testapp):
     }
 
 
-async def test_post_snippet(testapp, testconf):
+@pytest.mark.parametrize('snippet, rv', [
+    ({'content': 'def foo(): pass'},
+     {'id': 1,
+      'title': None,
+      'content': 'def foo(): pass',
+      'syntax': None,
+      'tags': [],
+      'created_at': _pytest_regex('\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}'),
+      'updated_at': _pytest_regex('\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}')}),
+
+    ({'title': 'snippet #1',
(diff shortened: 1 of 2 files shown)


ikalnytskyi:
> Makes sense: at the very least we'll be able to distinguish between cases when we stored `text`, because the syntax value wasn't passed vs when `text` was actually passed.

Yeah. Actually, I don't have strong preference between options (a) and (b), or maybe even some other. I think the main thing here is to be consistent and transparent in what we're doing. So if you option (a) is better, let me know.

Also, I'd like to ask you: do you think it's ok that we allow to do not pass `syntax` when we have constraint on allowed values? 

> Will xsnippet-web handle this correctly?

I believe snippet-web currently do not set "null" if syntax is not selected, as we always select at least `text` option. As for rendering, `null` should be treated as `text` syntax - no highlighting.

malor:
> Also, I'd like to ask you: do you think it's ok that we allow to do not pass syntax when we have constraint on allowed values?

I think having an additional "not set" state is ok.

> I believe snippet-web currently do not set "null" if syntax is not selected, as we always select at least text option. As for rendering, null should be treated as text syntax - no highlighting.

If that's already the case, then I think we are good to go.

PR #82, opened 2018-04-16
Version API via URL instead of HTTP header

ikalnytskyi:
Despite being deployed to production we are still in BETA, so we can and
probably should apply such changes especially because they were
discussed long ago.

So why this change? Well, API versioning via URL suits us better
because of the following reasons:

 * API users are forced to specify API version they want to use which
   illuminates ambiguousness we previously had when no version is
   supplied.

 * Troubleshooting is easier because we do not need to specially change
   logging format to show which version was used for URL is something
   that's logged either way.

 * Using API via cURL and co. is easier with passing version via URL.

 * We do not plan to dramatically change our API (only extend) which
   means we'll probably have v1 forever.

 * Versioning via URL does not require writing custom implementations
   of aiohttp abstractions we previously had. So less code - less bugs.


malor:
> However, if you think we better preserve old URLs for a while, I can add them back. Should I?

I guess I'd try to preserve the existing URLs.

ikalnytskyi:
> I guess I'd try to preserve the existing URLs.

For how long? Forever?

malor:
@ikalnytskyi Not necessarily, for a few releases maybe. We should probably also reflect the fact of backward-incompatible changes in the version number.

ikalnytskyi:
@malor 
> We should probably also reflect the fact of backward-incompatible changes in the version number.

I'm not sure it makes sense to expose backward incompatible changes via application version. I mean it's about breaking API, not application.. usually such things are done via bumping API version. Besides, we didn't even release xsnippet-api on pypi.

ikalnytskyi:
@malor I updated the PR. Please take a look one more time when you got a minute. ;)

xsnippet/api/routes.py
```diff
@@ -6,6 +6,14 @@
 
 
 v1 = [
+    web.route('*', '/v1/snippets', resources.Snippets),
+    web.route('*', '/v1/snippets/{id}', resources.Snippet),
+    web.route('*', '/v1/syntaxes', resources.Syntaxes),
+
+    # These routes are what we had before during era of API versioning through
```
malor: 👍

PR #88, opened 2018-05-04
Use envvar based configuration

ikalnytskyi:
XSnippet API does not provide a lot of configuration options, hence
there's a little sense in storing the config in external INI file.
Besides, since we use Docker and Docker Swarm to deploy production
environment, it'd be easier to parametrize deployment using environment
variables rather than commit changes to external file on filesystem.


xsnippet/api/conf.py
```diff
@@ -12,23 +12,24 @@
...
-
     :param envvar: an environment variable that points to additional config
     :type envvar: str
 
-    :return: a configuration instance
-    :rtype: :class:`configparser.ConfigParser`
+    :return: a configuration dictionary
+    :rtype: :class:`dict`
     """
+    # TODO: Drop INI support and use configuration via environment vars only.
```
ikalnytskyi: Please, notice this. :) The code of this function is terrible but it's a temporary measure to maintain backward compatibility for a while.

PR #95, opened 2018-07-08
Update travis ci config to test python 3.7 as well

malor:


malor:
The version of motor we use currently does not work on Python 3.7. The fix is merged to master (https://github.com/mongodb/motor/commit/9a88d92dadf998cffd657acdf77ab5e6d80596aa), but has not been released yet.

I guess, we can put this on hold for a little bit.

PR #96, opened 2018-09-27
Use built-in find_namespace_packages function

rkiyanchuk:
Originally setuptools.find_packages() could not handle PEP-420 namespace
packages, however since [v40.1.0][1] a new
function find_namespace_packages() is provided that is compliant with
PEP-420. Documentation for the function can be found [here][2].

[1]: https://github.com/pypa/setuptools/blob/master/CHANGES.rst#v4010
[2]: https://setuptools.readthedocs.io/en/latest/setuptools.html#find-namespace-packages

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@
 # coding: utf-8
 
 import os
-from setuptools import setup, find_packages as _find_packages
+from setuptools import setup, find_namespace_packages
 
 from xsnippet.api import __version__ as version
 from xsnippet.api import __license__ as license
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -32,7 +25,7 @@ def find_packages(namespace):
     keywords='web-service restful-api snippet storage',
     author='The XSnippet Team',
     author_email='dev@xsnippet.org',
-    packages=find_packages('xsnippet'),
+    packages=find_namespace_packages(include=['xsnippet.*']),
```
ikalnytskyi: Finally they made it! :+1:

ikalnytskyi:
We are going to rewrite it in Rust!

rkiyanchuk:
"Closed. Won't fix" again... 😒 
😄

PR #100, opened 2020-07-25
Initial database schema

malor:
Diesel is used for schema migrations (written in raw SQL; targetting
PostgreSQL at the moment) and conversions between SQL and Rust
data types.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,5 +16,7 @@ categories = [
 publish = false
 
 [dependencies]
+chrono = "0.4.13"
+diesel = { version = "1.4.5", features = ["postgres"] }
 rocket = "0.4.5"
 rocket_contrib = {version = "0.4.5", features = ["json"]}
--- a/src/storage/sql/migrations/2020-07-25-065258_create_snippets/up.sql
+++ b/src/storage/sql/migrations/2020-07-25-065258_create_snippets/up.sql
@@ -1,0 +1,59 @@
+CREATE TABLE snippets (
+    -- an internal autoincrementing identifier used in foreign keys.
+    -- Normally not visible publicly, except for the case when it is
+    -- used for looking legacy snippets up by id
+    id SERIAL PRIMARY KEY,
+    -- a short unique snippet identifier visible to users
+    slug VARCHAR(32) NOT NULL,
+
+    title TEXT,
+    syntax TEXT,
+    created_at TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL,
+    updated_at TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL,
+
+    -- slugs must be unique (this will also automatically create a unique index)
+    CONSTRAINT uq_slug UNIQUE (slug)
+);
+
+
+-- will be used for pagination; slug guarantees uniqueness of the sorting key
+CREATE INDEX snippets_created_at_slug ON snippets (created_at, slug);
+CREATE INDEX snippets_updated_at_slug ON snippets (updated_at, slug);
+
+
+CREATE TABLE changesets (
+    id SERIAL PRIMARY KEY,
+    snippet_id INTEGER NOT NULL,
+
+    -- numeric index used to determine the ordering of changesets for a given snippet
+    version INTEGER DEFAULT 0 NOT NULL,
+    content TEXT NOT NULL,
+    created_at TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL,
+    updated_at TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL,
+
+    -- there can be multiple changesets per snippet
+    CONSTRAINT fk_snippet FOREIGN KEY (snippet_id) REFERENCES snippets(id),
+    -- but each one is supposed to have a unique version number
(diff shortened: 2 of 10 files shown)


src/storage/sql/migrations/2020-07-25-065258_create_snippets/up.sql
```diff
@@ -0,0 +1,58 @@
...
+    -- Normally not visible publicly, except for the case when it is
+    -- used for looking legacy snippets up by id
+    id SERIAL PRIMARY KEY,
+    -- a short unique snippet identifier visible to users
+    slug TEXT NOT NULL,
+
+    title TEXT,
+    syntax TEXT,
+    created_at TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL,
+    updated_at TIMESTAMP WITH TIME ZONE,
```
ikalnytskyi, line 12: *(nitpicking)* Let's add `DEFAULT NOW() NOT NULL` here too. I feel like it might simplify some queries such as if we want to show some activity feed in the future.

src/storage/sql/migrations/2020-07-25-065258_create_snippets/up.sql
```diff
@@ -0,0 +1,58 @@
...
+
+CREATE TABLE changesets (
+    id SERIAL PRIMARY KEY,
+    snippet_id INTEGER NOT NULL,
+
+    -- numeric index used to determine the ordering of changesets for a given snippet
+    version INTEGER DEFAULT 0 NOT NULL,
+    content TEXT NOT NULL,
+    created_at TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL,
+    updated_at TIMESTAMP WITH TIME ZONE,
```
ikalnytskyi, line 31: *(nitpicking)* Let's add `DEFAULT NOW() NOT NULL` here too. I feel like it might simplify some queries such as if we want to show some activity feed in the future.

src/storage/sql/migrations/2020-07-25-065258_create_snippets/up.sql
```diff
@@ -0,0 +1,58 @@
+CREATE TABLE snippets (
+    -- an internal autoincrementing identifier used in foreign keys.
+    -- Normally not visible publicly, except for the case when it is
+    -- used for looking legacy snippets up by id
+    id SERIAL PRIMARY KEY,
+    -- a short unique snippet identifier visible to users
+    slug TEXT NOT NULL,
```
ikalnytskyi, line 7: **Q:** Should we limit `slug` by, let's say, 100 characters or use some `VARCHAR` instead? Or you think `TEXT` is simpler and preferable in our case?

src/storage/sql/migrations/00000000000000_diesel_initial_setup/up.sql
```diff
@@ -0,0 +1,36 @@
...
+CREATE OR REPLACE FUNCTION diesel_set_updated_at() RETURNS trigger AS $$
+BEGIN
+    IF (
+        NEW IS DISTINCT FROM OLD AND
+        NEW.updated_at IS NOT DISTINCT FROM OLD.updated_at
+    ) THEN
+        NEW.updated_at := current_timestamp;
+    END IF;
+    RETURN NEW;
+END;
```
ikalnytskyi, line 35: Do I understand correctly that with this we won't need to manage `updated_at` manually from the code?

src/storage/sql/migrations/2020-07-25-065258_create_snippets/up.sql
malor: Agreed. Done.

src/storage/sql/migrations/2020-07-25-065258_create_snippets/up.sql
malor: Agreed. Done.

src/storage/sql/migrations/2020-07-25-065258_create_snippets/up.sql
malor: AFAIK, it makes no difference for efficiency, but we could use VARCHAR(N) as a sanity check. I think, 100 is way too high, though (as long we use both lowecase and uppercase letters + 0-9 digits in our range of allowed symbols).

src/storage/sql/migrations/00000000000000_diesel_initial_setup/up.sql
malor: Yes, that's the idea. Actually, let me omit this for now, and bring it back if it turns out to be useful.

Cargo.toml
```diff
@@ -16,5 +16,7 @@ categories = [
 publish = false
 
 [dependencies]
+chrono = "0.4.13"
+diesel = { version = "1.4.5", features = ["postgres"] }
```
ikalnytskyi, line 20: **Q:** Do we need to install `diesiel-cli` from `master`?

malor: My current understanding is that this is not possible at the moment, because `diesel_cli` is a binary-only crate :( https://github.com/rust-lang/cargo/issues/4316

Let me try using https://docs.rs/diesel_migrations/1.4.0/diesel_migrations/index.html in a follow up PR. Or just replace this with alembic ;)

PR #102, opened 2020-07-26
Use alembic for schema migrations

malor:
Pros:

* we do not have to write raw SQL; alembic allows us to target more
than one database backend using the same code (e.g. so that we can
run tests on SQLite locally)

* the directory with migration scripts can be placed anywhere within
the repo, unlike with diesel_cli that has not been released with
this feature yet :)

* alembic is more mature and feature-rich

Cons:

* one more tool in a different programming language (Python)

* one more step to generate Diesel schema macros (diesel_cli combines
this step with running of the migration scripts). We will need to run
the following command every time we change the database schema:

```
    $ diesel print-schema > src/storage/sql/schema.rs
```

* we won't have the option of not shipping the migration scripts with
the code

Usage:

```
$ python -m pip install alembic psycopg2
$ export DATABASE_URL=postgres://user:password@host:port/database
$ export DATABASE_URL=sqlite:///path/to/db
$ alembic upgrade head  # upgrade to the latest revision
$ alembic downgrade base  # downgrade to the base (empty) revision
$ alembic current  # print the current revision
```


.github/workflows/tests.yml
```diff
@@ -48,13 +48,46 @@ jobs:
...
         run: |
-          choco install postgresql12 --force --params '/Password:root'
+          choco install postgresql12 --force
           echo '::add-path::C:\Program Files\PostgreSQL\12\bin'
           echo '::add-path::C:\Program Files\PostgreSQL\12\lib'
+          echo '::set-env name=PQ_LIB_DIR::C:\Program Files\PostgreSQL\12\lib'
+          "C:\Program Files\PostgreSQL\12\bin\createuser" -s devel
+          "C:\Program Files\PostgreSQL\12\bin\createdb" -O devel devel
+          "C:\Program Files\PostgreSQL\12\bin\psql" -c "ALTER USER devel PASSWORD 'devel';" devel
+          echo '::set-env name=DATABASE_URL::postgres://devel:devel@localhost/devel'
```
ikalnytskyi, line 81: **Q:** I wonder, would it make sense to consider using `services` feature [1] to spin up a postgresql container and avoid duplicate commands for every OS we support?  

[1] https://docs.github.com/en/actions/configuring-and-managing-workflows/creating-postgresql-service-containers#running-jobs-directly-on-the-runner-machine

src/storage/sql/migrations/versions/1bf4e4e7b24f_initial_schema.py
```diff
@@ -0,0 +1,103 @@
...
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        'snippets',
+
+        # an internal autoincrementing identifier used in foreign keys.
+        # Normally not visible publicly, except for the case when it is
+        # used for looking legacy snippets up by id
```
ikalnytskyi, line 25: I was actually thinking that for legacy snippets we can set `slug` to be their old `id` and avoid having special logic in the code base. What do you think?

src/storage/sql/migrations/versions/1bf4e4e7b24f_initial_schema.py
```diff
@@ -0,0 +1,103 @@
...
+        # sanity check: version numbers are non-negative integers
+        sa.CheckConstraint('version >= 0',
+                           name='check_non_negative_version')
+    )
+    with op.batch_alter_table('changesets') as batch_op:
+        # there can be multiple changesets per snippet; changesets should be
+        # deleted when the parent snippet is deleted
+        batch_op.create_foreign_key('fk_snippet', 'snippets',
+                                    ['snippet_id'], ['id'],
+                                    ondelete='CASCADE')
```
ikalnytskyi, line 75: **Q:** For my education, can you please explain why `batch_alter_table` is used? Can't we just create a foreign key sing `op.` or even as a part of `op.create_table` ?

src/storage/sql/migrations/versions/1bf4e4e7b24f_initial_schema.py
```diff
@@ -0,0 +1,103 @@
...
+    with op.batch_alter_table('tags') as batch_op:
+        # there can be multiple tags per snippet; tags should be deleted when
+        # the parent snippet is deleted
+        batch_op.create_foreign_key('fk_snippet', 'snippets',
+                                    ['snippet_id'], ['id'],
+                                    ondelete='CASCADE')
+
+
+def downgrade():
+    op.drop_table('tags')
```
ikalnytskyi, line 101: Just to double confirm, you decided that we don't want to go with intermediate `snippets_tags` table that glues together `tags` and  `snippets`?

src/storage/sql/migrations/versions/1bf4e4e7b24f_initial_schema.py
malor: Oh, that's an interesting point. We could treat them as strings, indeed! I want to keep the integer IDs, though, as they are a better fit for foreign key constraints (because they are smaller and, thus, the indexes are smaller, and also comparing 4-byte integers is likely slightly more efficient).

src/storage/sql/migrations/versions/1bf4e4e7b24f_initial_schema.py
malor: This is the first time I've used this as well. The short answer is that this is only needed to support SQLite :) You can't ALTER tables in SQLite, so what Alembic does is that it creates a new table under the hood and copies all existing rows there. My understanding is that they restrict the interface of  `create_foreign_key()` to `batch_alter_table()` specifically so that all modifications are batched and the copy is only performed once.

src/storage/sql/migrations/versions/1bf4e4e7b24f_initial_schema.py
malor: Correct. In order to keep things simpler on the querying side, as apparently Diesel DSL is a bit picky about multiple JOINs in the same query.

src/storage/sql/migrations/versions/1bf4e4e7b24f_initial_schema.py
ikalnytskyi: Hm, I think I'm still missing the point. Are you saying that if I run this migration with SQLlite engine, there will be no `op.create_foreign_key`? Also, even if SQLite does not support ALTER, can't we define a first time foreign key constraint in the same SQL query that creates a table?

.github/workflows/tests.yml
malor: As discussed, that's only possible on Linux at the moment :( The docs you pointed me to contained a valuable hint, though: PostgreSQL is already installed on Windows; it's just disabled. I will try to update this script to start the service instead of reinstalling it from scratch, which should shave off some running time.

src/storage/sql/migrations/versions/1bf4e4e7b24f_initial_schema.py
ikalnytskyi: Integer IDs are OK. :+1: Another question I have is what are we going to do with both `slug` and `id` in term of API response? Should we use `slug` as an `id` in API response? Or you want to explicitly separate these two for API users?

src/storage/sql/migrations/versions/1bf4e4e7b24f_initial_schema.py
malor: > Are you saying that if I run this migration with SQLlite engine, there will be no op.create_foreign_key? 

There will be. But it will fail at runtime and point you to `batch_alter_table()` docs.

> Also, even if SQLite does not support ALTER, can't we define a first time foreign key constraint in the same SQL query that creates a table?

We totally can. I just could not remember how to do it off the top of my head :) I'll update the code.

src/storage/sql/migrations/versions/1bf4e4e7b24f_initial_schema.py
malor: My plan was to treat integer IDs as an implementation detail and not expose them via the API. Perhaps, only put `slug` to the response under the name `id`?

PR #104, opened 2020-08-16
The initial implementation of the storage layer

malor:
Trait `Storage` defines a basic CRUD interface for storing/loading snippets from a persistent database. The only implementation planned as of now is `SqlStorage,` which only supports PostgreSQL at the moment, while in the future it should be relatively easy to extend it to support SQLite and MySQL if desired (we would need to stop using `INSERT ... ON CONFLICT DO ...` , though).

### DIFF
--- a/src/application.rs
+++ b/src/application.rs
@@ -2,6 +2,7 @@ use std::collections::HashSet;
 use std::error::Error;
 
 use super::routes;
+use super::storage::{SqlStorage, Storage};
 
 #[derive(Debug)]
 pub struct Config {
@@ -34,7 +35,14 @@ pub fn create_app() -> Result<rocket::Rocket, Box<dyn Error>> {
         Err(err) => return Err(Box::new(err)),
     };
 
+    let database_url = match app.config().get_string("database_url") {
+        Ok(database_url) => database_url,
+        Err(err) => return Err(Box::new(err)),
+    };
+    let storage: Box<dyn Storage> = Box::new(SqlStorage::new(&database_url)?);
+
     Ok(app
         .manage(Config { syntaxes })
+        .manage(storage)
         .mount("/v1", routes![routes::syntaxes::get_syntaxes]))
 }
--- a/src/storage/errors.rs
+++ b/src/storage/errors.rs
@@ -1,0 +1,63 @@
+use std::error;
+use std::fmt;
+
+/// All errors that can be returned by the storage layer wrapped into a single
+/// enum type for convinience.
+#[derive(Debug)]
+pub enum StorageError {
+    /// Snippet with this id already exists in the database
+    Duplicate { id: String },
+    /// Snippet with this id can't be found
+    NotFound { id: String },
+    /// All other errors that can't be handled by the storage layer
+    InternalError(Box<dyn error::Error>),
+}
+
+impl fmt::Display for StorageError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            StorageError::Duplicate { id } => write!(f, "Snippet with id `{}` already exists", id),
+            StorageError::NotFound { id } => write!(f, "Snippet with id `{}` is not found", id),
+            StorageError::InternalError(e) => write!(f, "Internal error: `{}`", e),
+        }
+    }
+}
+
+// this is not strictly required, but allows for wrapping a StorageError
+// instance into a Box<dyn Error> if needed. Error's only requirement is for
+// a type to implement Debug + Display
+impl error::Error for StorageError {}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_display() {
+        assert_eq!(
+            "Snippet with id `spam` already exists",
+            format!(
+                "{}",
+                StorageError::Duplicate {
+                    id: "spam".to_string()
+                }
+            )
+        );
+
+        assert_eq!(
+            "Snippet with id `eggs` is not found",
+            format!(
+                "{}",
+                StorageError::NotFound {
+                    id: "eggs".to_string()
+                }
+            )
+        );
+
+        let error = "foo".parse::<f32>().err().unwrap();
+        assert_eq!(
+            "Internal error: `invalid float literal`",
+            format!("{}", StorageError::InternalError(Box::new(error)))
+        );
+    }
+}
--- a/src/storage/models.rs
+++ b/src/storage/models.rs
@@ -1,0 +1,261 @@
+use std::iter;
+
+use chrono::{DateTime, Utc};
+use rand::Rng;
+
+const DEFAULT_SLUG_LENGTH: usize = 8;
+
+/// A code snippet
+#[derive(Debug, Default, Eq, PartialEq)]
+pub struct Snippet {
+    /// Slug that uniquely identifies the snippet
+    pub id: String,
+    /// Snippet title. None means that the title was not specified at snippet
+    /// creation time
+    pub title: Option<String>,
+    /// Snippet syntax. None means that the syntax was not specified at snippet
+    /// creation time
+    pub syntax: Option<String>,
+    /// List of revisions. The snippet content is the content of the most recent
+    /// revision. Revisions are sorted by the version number in the
+    /// ascending order.
+    pub changesets: Vec<Changeset>,
+    /// List of tags attached to the snippet
+    pub tags: Vec<String>,
+    /// Timestamp of when the snippet was created
+    pub created_at: Option<DateTime<Utc>>,
+    /// Timestamp of when the snippet was last modified. Defaults to creation
+    /// time
+    pub updated_at: Option<DateTime<Utc>>,
+}
+
+impl Snippet {
+    /// Create a new Snippet.
+    pub fn new(
+        title: Option<String>,
+        syntax: Option<String>,
+        changesets: Vec<Changeset>,
+        tags: Vec<String>,
+    ) -> Self {
+        let mut snippet = Snippet::default();
+        snippet.id = Snippet::random_id(DEFAULT_SLUG_LENGTH);
+        snippet.title = title;
+        snippet.syntax = syntax;
+        snippet.changesets = changesets;
+        snippet.tags = tags;
+
+        snippet
+    }
+
+    /// Generate a random unique snippet identifier (slug).
+    pub fn random_id(length: usize) -> String {
+        let mut rng = rand::thread_rng();
+        iter::repeat_with(|| rng.sample(rand::distributions::Alphanumeric))
+            .take(length)
+            .collect()
+    }
+}
+
+/// A particular snippet revision
+#[derive(Debug, Default, Eq, Ord, PartialEq, PartialOrd)]
+pub struct Changeset {
+    /// Changeset index. Version numbers start from 0 and are incremented by 1
+    pub version: usize,
+    /// Changeset content
+    pub content: String,
+    /// Timestamp of when the changeset was created
+    pub created_at: Option<DateTime<Utc>>,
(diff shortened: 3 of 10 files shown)


src/application.rs
```diff
@@ -34,7 +35,14 @@ pub fn create_app() -> Result<rocket::Rocket, Box<dyn Error>> {
         Err(err) => return Err(Box::new(err)),
     };
 
+    let database_url = match app.config().get_string("database_url") {
+        Ok(database_url) => database_url,
+        Err(err) => return Err(Box::new(err)),
+    };
+    let storage: Box<dyn Storage> = Box::new(SqlStorage::new(&database_url)?);
```
ikalnytskyi, line 42: **Q:** What is `database_url`? Is it `String` or `&str`?

src/storage/errors.rs
```diff
@@ -0,0 +1,63 @@
...
+            )
+        );
+
+        let error = "foo".parse::<f32>().err().unwrap();
+        assert_eq!(
+            "Internal error: `invalid float literal`",
+            format!("{}", StorageError::InternalError(Box::new(error)))
+        );
+    }
+}
```
ikalnytskyi, line 63: Kek :laughing:

src/storage/models.rs
```diff
@@ -0,0 +1,261 @@
...
+
+impl Snippet {
+    /// Create a new Snippet.
+    pub fn new(
+        title: Option<String>,
+        syntax: Option<String>,
+        changesets: Vec<Changeset>,
+        tags: Vec<String>,
+    ) -> Self {
+        let mut snippet = Snippet::default();
```
ikalnytskyi, line 40: **Q:** What `::default()` is? Where is it implemented?

src/storage/models.rs
```diff
@@ -0,0 +1,261 @@
...
+        changeset.version = version;
+        changeset.content = content;
+
+        changeset
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
```
ikalnytskyi, line 86: TBH, I'm not a fan of that Rust idea of writing test in the same module where the code lies. :-/

src/storage/sql/mod.rs
```diff
@@ -1 +1,319 @@
...
+        Ok(())
+    }
+}
+
+impl Storage for SqlStorage {
+    fn create(&self, snippet: &Snippet) -> Result<Snippet, StorageError> {
+        let conn = self.pool.get()?;
+        conn.transaction::<_, StorageError, _>(|| {
+            // insert the new snippet row first to get the generated primary key
+            let snippet_id = self.upsert_snippet(&conn, snippet)?;
```
ikalnytskyi, line 133: **Q:** If it's *upsert*, does it mean it either updates or inserts the record? If so, shouldn't we forbid updating here? Especially when ID conflict is araised?

src/storage/sql/mod.rs
```diff
@@ -1 +1,319 @@
...
+        } else {
+            // otherwise, potentially update the title and the syntax
+            let conn = self.pool.get()?;
+            conn.transaction::<_, StorageError, _>(|| {
+                let snippet_id = self.upsert_snippet(&conn, snippet)?;
+                // insert new changesets and tags
+                self.upsert_changesets(&conn, snippet_id, snippet)?;
+                self.upsert_tags(&conn, snippet_id, snippet)?;
+                // and delete the removed tags
+                self.trim_removed_tags(&conn, snippet_id, snippet)?;
```
ikalnytskyi, line 184: **Q:** Maybe we should think of grouping these together?

src/storage/sql/mod.rs
```diff
@@ -1 +1,319 @@
...
+        // load the snippet from the db to check if we need to update anything
+        let persisted_state = self.get(&snippet.id)?;
+        if persisted_state == *snippet {
+            // if not, simply return the current state
+            Ok(persisted_state)
+        } else {
+            // otherwise, potentially update the title and the syntax
+            let conn = self.pool.get()?;
+            conn.transaction::<_, StorageError, _>(|| {
+                let snippet_id = self.upsert_snippet(&conn, snippet)?;
```
ikalnytskyi, line 179: **Q:** What if there wasn't a snippet in the database? Should `update` implicitly create one?

ikalnytskyi (approved):
Good work! I haven't checked tests though.  :)

src/application.rs
malor: `String`. That's why I need to add a `&` when passing it to `SqlStorage::new()`

src/storage/errors.rs
malor: Yeah, it was the first internal error I could think of :)

src/storage/models.rs
malor: It is a method of the `Default` trait (https://doc.rust-lang.org/std/default/trait.Default.html).

The implementation varies from type to type. E.g. for `usize` it will return `0`, for `f32` it will return `0.0`, etc. For structs you can derive the trait `Default` if every field of the struct implements it.

We don't necessarily *need* it, and can either always use the `Snippet::new()` method, or initialize all the fields explicitly. But, I think, it may be handy for creating an "empty" Snippet instance, while `Snippet::new()` produces an instance with all the required fields populated.

src/storage/models.rs
malor: We totally could put this into a separate module and include it here (so that the test module would have access to the private functions/structs/fields of this one), but that would not be idiomatic, AFAIK :(

src/storage/sql/mod.rs
malor: >  If it's upsert, does it mean it either updates or inserts the record?

Correct. It tries to `INSERT` a new row first, and falls back to `UPDATE` in case of a *conflict* (violation of some form of a unique constraint).

>  If so, shouldn't we forbid updating here? Especially when ID conflict is araised?

That's a very good observation! :) As is, the code assumes that duplicate slugs never happen (after all, the probability of such event is tiny: `1 / len(string.ascii_letters + string.digits) ** 8 = 4.580010609233563e-15`), so you are right that an upsert would not work for detection of conflicts, as we simply `UPDATE` the existing row in-place in case of a conflict.

I'm not sure what we should do here :( What I like about the current approach is that it is concise and allows to reuse the very same code for both new insertions and updates. But, as you correctly pointed out, this rules out our initial idea of generating longer IDs on retries.

src/storage/sql/mod.rs
malor: It doesn't at the moment, but it totally could, because we share the code for `update()` and `create()`.

I guess, it's a question of whether we want to give a signal to the caller when an update is issued for a non-existing snippet or not. What would be your preference?

src/storage/sql/mod.rs
malor: Not sure I understand what you mean by that. Grouping what?

malor (commented):
Thank you for the comments! ;)

src/storage/sql/mod.rs
ikalnytskyi: > I'm not sure what we should do here :(

Neither do I . I wonder, what others do in this case? :thinking: One thing is clear for me, we definitely need to drop a comment here explaining a possible gotcha. At least it will show others (potential readers and/or contributors), that it's not a bug but an intentional move.

src/storage/sql/mod.rs
ikalnytskyi: I was thinking that `POST` will directly map to `create()`, while `PUT` -- to `update()`. According to widespread practice, `PUT` must receive the whole resource representation (not partial, hence all non provided properties will be reset to defaults), and **can** create a resource if it wasn't here before. The only requirement here is that the `id` must be provided beforehand (i.e. server-side generation must not be involved). For partial updates, `PATCH` must be used instead.

So I think we have two questions here:

1. Do we want to support both `PUT` and `PATCH`, as the REST purists would prefer?
2. Do we want to allow users to provide their own `id`? 

Re 1: I tend to believe it might be useful to have both `PUT` and `PATCH`, as long as it's not hard to implement.
Re 2: If we let users to define their own `id`, we definitely need to think about ID duplicates in `create()` and `update()` because it would be much easier to hit this rare case.

WDYT?

malor: > Re 1: I tend to believe it might be useful to have both PUT and PATCH, as long as it's not hard to implement.

Agreed. That sounds good to me!

>  Re 2: If we let users to define their own id, we definitely need to think about ID duplicates in create() and update() because it would be much easier to hit this rare case.

Yeah, good point! TBH, I'd rather not allow users to choose snippet identifiers. I can't think of a specific a way of how they could potentially exploit that, but that sounds like giving too much control and an opportunity for name squatting.

src/storage/sql/mod.rs
malor: That's fair! I'll update the code.

malor: Split it into two separate methods, so that we can properly detect a unique key violation.

src/storage/sql/mod.rs
```diff
@@ -1 +1,319 @@
...
+            .get_result::<i32>(conn);
+        let snippet_id = match result {
+            Ok(snippet_id) => snippet_id,
+            Err(DatabaseError(DatabaseErrorKind::UniqueViolation, _)) => {
+                return Err(StorageError::Duplicate {
+                    id: snippet.id.to_owned(),
+                })
+            }
+            Err(e) => return Err(StorageError::from(e)),
+        };
```
ikalnytskyi, line 52: If it's `upsert`, in which case it may return a duplicate error?

src/storage/sql/mod.rs
```diff
@@ -1 +1,319 @@
...
+
+        Ok(snippet_id)
+    }
+
+    fn upsert_changesets(
```
ikalnytskyi, line 57: It's not public, isn't it? I mean, for outer world changesets must be immutable.

src/storage/sql/mod.rs
ikalnytskyi: > `let persisted_state = self.get(&snippet.id)?;`

Oh, I see. It will return an error if there's no such snippet in the database. I somehow didn't notice that trailing `?` question mark. I think we're good here.

src/storage/sql/mod.rs
ikalnytskyi: Just ignore that.

ikalnytskyi (approved):

src/storage/sql/mod.rs
malor: Ok :)

src/storage/sql/mod.rs
malor: Correct!

ikalnytskyi (approved):

src/storage/sql/mod.rs
```diff
@@ -1 +1,330 @@
...
+        Ok(Self { pool })
+    }
+
+    fn insert_snippet(&self, conn: &PgConnection, snippet: &Snippet) -> Result<i32, StorageError> {
+        let result = diesel::insert_into(snippets::table)
+            .values((
+                snippets::slug.eq(&snippet.id),
+                snippets::title.eq(&snippet.title),
+                snippets::syntax.eq(&snippet.syntax),
+                snippets::created_at.eq(diesel::dsl::now),
```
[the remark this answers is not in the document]
malor: Ah, good catch! We *do* have a server default for both `created_at` and `updated_at` set to the value of `NOW()`, but I agree that this should be consistent, and we should either set both in the code, or none of them.

src/storage/sql/mod.rs
malor: Sorry, this I forgot to push the updated commit :(

PR #105, opened 2021-01-10
Add declarative HTTP tests using Gabbi

ikalnytskyi:
Gabbi is a tool for running HTTP tests where requests and responses are
expressed as declarations in YAML files [1]. We want to use this
framework to end-to-end XSnippet API's functionality, and make tests as
language/framework agnostic as possible. With Gabbi in-place, it should
be possible with little changes to run these tests against any
implementation of our API.

[1] https://gabbi.readthedocs.io


tests/test_gabbits.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def __init__(self):
+        self.environ = {
+            "ROCKET_HOST": XSNIPPET_API_HOST,
+            "ROCKET_PORT": str(XSNIPPET_API_PORT),
+            "ROCKET_DATABASE_URL": XSNIPPET_API_DATABASE_URL,
+        }
+
+        if self._syntaxes:
+            # ROCKET_SYNTAXES expects a TOML array as input. Since TOML library
```
malor, line 36: Heh :)

tests/test_gabbits.py
```diff
@@ -0,0 +1,107 @@
...
+
+        self.process = None
+
+    def start_fixture(self):
+        """Start the live server."""
+
+        environ = os.environ.copy()
+        environ.update(self.environ)
+
+        self.process = subprocess.Popen(self._launch_command, env=environ, shell=True)
```
malor, line 50: nit: passing `shell=True` adds a middleman between the test process and the xsnippet api running in the background, so the termination signal will have to go through shell process -> cargo process -> xsnippet api process. If termination works correctly now, then both shell and cargo forward the signals correctly, though, so maybe it's not a big deal.

tests/test_gabbits.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def _wait_for_socket(host, port, timeout):
+    """Wait for socket to start accepting connections."""
+
+    start_time = time.monotonic()
+
+    while True:
+        try:
+            with socket.create_connection((host, port), timeout=timeout):
```
malor, line 89: TIL about socket.create_connection() :+1:

tests/test_gabbits.py
```diff
@@ -0,0 +1,107 @@
...
+        return syntaxes
+
+
+class XSnippetApiWithNoSyntaxes(XSnippetApi):
+    """Start live server of XSnippet API with no syntaxes configured."""
+
+    _syntaxes = None
+
+
+def _wait_for_socket(host, port, timeout):
```
malor, line 82: This is perfectly fine for now. In the future we might want to add some HTTP endpoint that will allow us to verify that the application has been initialized correctly and the service is healthy.

malor (approved):
Nice! You just need to rebase.

tests/test_gabbits.py
ikalnytskyi: That's a nice point. I initially started with "shell=False" and changed it to `True` along the way trying to overcome various issue.

PR #108, opened 2021-01-15
Add 'setup-postgres' action

ikalnytskyi:
In order to improve workflows' readability, let's move the steps that
setup postgresql server into standalone action. This way it will be
easier reuse this code between the jobs, and the jobs will be much
simpler.

### DIFF
--- a/.github/actions/setup-postgres/action.yml
+++ b/.github/actions/setup-postgres/action.yml
@@ -1,0 +1,62 @@
+name: Setup PostgreSQL
+description: |
+  Setup a PostgreSQL for Linux, macOS and Windows machines. Runs natively, no
+  Docker containers are involved.
+inputs:
+  username:
+    description: A username of the user to setup.
+    default: postgres
+  password:
+    description: A password of the user to setup.
+    default: postgres
+  database:
+    description: A database name to setup and grant permissions to created user.
+    default: postgres
+runs:
+  using: composite
+  steps:
+    - name: Prerequisites
+      run: |
+        if [ "$RUNNER_OS" == "Linux" ]; then
+          echo "$(pg_config --bindir)" >> $GITHUB_PATH
+        elif [ "$RUNNER_OS" == "Windows" ]; then
+          echo "$PGBIN" >> $GITHUB_PATH
+          echo "PQ_LIB_DIR=$PGROOT\lib" >> $GITHUB_ENV
+        fi
+      shell: bash
+
+
+    - name: Setup and start PostgreSQL
+      run: |
+        export PGDATA="$RUNNER_TEMP/pgdata"
+        pg_ctl init
+
+        # Forbid creating unix sockets since they are created by default in the
+        # directory we don't have permissions to.
+        echo "unix_socket_directories = ''" >> "$PGDATA/postgresql.conf"
+        pg_ctl start
+
+        # Both PGHOST and PGUSER are used by PostgreSQL tooling such as 'psql'
+        # or 'createuser'. Since PostgreSQL data has been resetup, we cannot
+        # rely on defaults anymore.
+        #
+        # PGHOST is required for Linux and macOS since they default to unix
+        # sockets, and we have turned them off.
+        #
+        # PGUSER is required for Windows since default the tooling's default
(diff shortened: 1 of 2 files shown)


.github/actions/setup-postgres/action.yml
```diff
@@ -0,0 +1,62 @@
...
+          echo "$PGBIN" >> $GITHUB_PATH
+          echo "PQ_LIB_DIR=$PGROOT\lib" >> $GITHUB_ENV
+        fi
+      shell: bash
+
+
+    - name: Setup and start PostgreSQL
+      run: |
+        export PGDATA="$RUNNER_TEMP/pgdata"
+        pg_ctl init
```
malor, line 32: Do you know if there is an easy way for us to check the contents of `$PGDATA/postgresql.conf` at this point? (maybe just `cat` it to stdout?). I wonder what `listen_addresses` is set to (https://www.postgresql.org/docs/13/runtime-config-connection.html#GUC-LISTEN-ADDRESSES). Now that we've disabled unix sockets, it would be nice to double check that we do not allow remote connections and only listen on 127.0.0.1 and/or ::1. Or am I being paranoid and Github Actions already enforce strict network isolation?

.github/actions/setup-postgres/action.yml
```diff
@@ -0,0 +1,62 @@
...
+        # PGUSER is required for Windows since default the tooling's default
+        # user is 'postgres', while 'pg_ctl init' creates one with the name of
+        # the current user.
+        echo "PGHOST=localhost" >> $GITHUB_ENV
+        echo "PGUSER=${USER:-$USERNAME}" >> $GITHUB_ENV
+      shell: bash
+
+    - name: Setup PostgreSQL user and database
+      run: |
+        createuser -s ${{ inputs.username }}
```
malor, line 55: nit: do we need this user to be a superuser? I can't think of how this can potentially be exploited, though, so it's more of "the principle of least privilege" type of comment, rather than an actual concern :)

malor (approved):
Thank you for cleaning this up! 👍

.github/actions/setup-postgres/action.yml
ikalnytskyi: I'm not sure what default is, it could be 0.0.0.0. You should be able to see the listening address in logs since Postgres prints them on start.

I could explicitly set it to 127.0.0.1 though.

.github/actions/setup-postgres/action.yml
ikalnytskyi: Dunno. I just copied this over from before. I was hoping to make this action standalone and move into re-usable separate repository. Do you think it'd make sense to add another input Boolean parameter to control whether a user should be created as superuser?

.github/actions/setup-postgres/action.yml
ikalnytskyi: From logs:

```
2021-01-16 22:58:27.988 UTC [3484] LOG:  listening on IPv6 address "::1", port 5432
2021-01-16 22:58:27.989 UTC [3484] LOG:  listening on IPv4 address "127.0.0.1", port 5432
```

BTW, even if unix sockets are enabled, tcp/ip sockets are there anyway. My understanding is that tooling such as `createuser` and `psql` were using unix sockets by default, and the path to these sockets has been set compile time (unless overriden via `PGHOST` env var).

malor: Thank you for confirming! :+1: 

> BTW, even if unix sockets are enabled, tcp/ip sockets are there anyway. My understanding is that tooling such as createuser and psql were using unix sockets by default, and the path to these sockets has been set compile time (unless overriden via PGHOST env var).

Right. It can get confusing at times, as they will also use unix sockets when you use `localhost` in the DB connection string.

.github/actions/setup-postgres/action.yml
malor: I don't think we really need a super user, just a basic role that owns one particular database and is allowed create new tables there. Anyway, it's not a big deal if this is run in a VM, and Postgresql only listens on the loopback interface.

PR #109, opened 2021-01-17
Make Snippet and Changeset serializable

malor:
In order to return Snippets in API responses, we need to make them serializable first. Rocket can handle serialization to JSON, but it requires the types to implement `serde::Serializable`.

In theory, we could just auto-derive that trait for our structs, but in practice it is more complicated than that, as `created_at` and `updated_at` are of type `DateTime<Utc>`, and the latter does not implement `serde::Serializable`.

Rust has this fair, but sometimes annoying requirement, that you can't implement foreign traits for foreign types. So the way to work around that is to create a wrapper struct around `DateTime<Utc>` from the chrono crate. This has no runtime/memory overhead, but it is a little verbose (because we need to handle type conversions).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,3 +21,7 @@ diesel = { version = "1.4.5", features = ["chrono", "postgres", "r2d2"] }
 rand = "0.7.3"
 rocket = "0.4.5"
 rocket_contrib = {version = "0.4.5", features = ["json"]}
+serde = { version = "1.0", features = ["derive"] }
+
+[dev-dependencies]
+serde_json = "1.0"
--- a/src/storage/util.rs
+++ b/src/storage/util.rs
@@ -1,0 +1,64 @@
+use serde::{Serialize, Serializer};
+
+type ChronoUtc = chrono::DateTime<chrono::Utc>;
+
+/// A wrapper around DateTime<Utc> from chrono that allows us
+/// to implement Serialize (otherwise, Rust forbids implementing
+/// a foreign trait for a foreign type from a different crate).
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -21,3 +21,7 @@ diesel = { version = "1.4.5", features = ["chrono", "postgres", "r2d2"] }
 rand = "0.7.3"
 rocket = "0.4.5"
 rocket_contrib = {version = "0.4.5", features = ["json"]}
+serde = { version = "1.0", features = ["derive"] }
+
+[dev-dependencies]
+serde_json = "1.0"
```
malor, line 27: Rocket depends on serde_json. We are not using it directly in the production code, only for unit tests.

src/storage/util.rs
```diff
@@ -0,0 +1,64 @@
+use serde::{Serialize, Serializer};
+
+type ChronoUtc = chrono::DateTime<chrono::Utc>;
+
+/// A wrapper around DateTime<Utc> from chrono that allows us
+/// to implement Serialize (otherwise, Rust forbids implementing
+/// a foreign trait for a foreign type from a different crate).
```
ikalnytskyi, line 7: Rust does not have a built-in type for date type?

ikalnytskyi (approved):

src/storage/util.rs
malor: Nope :( chrono seems to be the most popular one.

ikalnytskyi: What a shame

ikalnytskyi: What a shame.

PR #112, opened 2021-01-23
Implement the POST /snippets route

malor:
This does the bare minimum to implement the API of XSnippet v4 as closely as possible.
    
One caveat is that the Gabbit tests can't clean up after themselves until DELETE is implemented as well.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,0 +1,88 @@
+use std::convert::From;
+use std::io::Cursor;
+
+use rocket::http;
+use rocket::request::Request;
+use rocket::response::{self, Responder, Response};
+use serde::{ser::SerializeStruct, Serialize, Serializer};
+
+use crate::storage::StorageError;
+
+/// All possible unsuccessful outcomes of an API request.
+///
+/// Allows to handle application errors in the unified manner:
+///
+/// 1) all errors are serialized to JSON messages like {"message": "..."};
+///    the HTTP status code is set accordingly
+///
+/// 2) implements conversions from internal errors types (e.g. the errors
+///    returned by the Storage trait)
+#[derive(Debug)]
+pub enum ApiError {
+    BadRequest(String),
+    NotFound(String),
+    InternalError(String),
+    UnsupportedMediaType(String),
+}
+
+impl ApiError {
+    /// Reason why the request failed.
+    pub fn reason(&self) -> &str {
+        match self {
+            ApiError::BadRequest(msg) => &msg,
+            ApiError::NotFound(msg) => &msg,
+            ApiError::InternalError(msg) => &msg,
+            ApiError::UnsupportedMediaType(msg) => &msg,
+        }
+    }
+
+    /// HTTP status code.
+    pub fn status(&self) -> http::Status {
+        match self {
+            ApiError::BadRequest(_) => http::Status::BadRequest,
+            ApiError::NotFound(_) => http::Status::NotFound,
+            ApiError::UnsupportedMediaType(_) => http::Status::UnsupportedMediaType,
+            ApiError::InternalError(_) => http::Status::InternalServerError,
+        }
+    }
+}
+
+impl From<StorageError> for ApiError {
+    fn from(value: StorageError) -> Self {
+        match value {
+            StorageError::NotFound { id: _ } => ApiError::NotFound(value.to_string()),
+            _ => ApiError::InternalError(value.to_string()),
+        }
+    }
+}
+
+impl Serialize for ApiError {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut s = serializer.serialize_struct("", 1)?;
+        s.serialize_field("message", self.reason())?;
+        s.end()
+    }
+}
+
+impl<'r> Responder<'r> for ApiError {
+    fn respond_to(self, _request: &Request) -> response::Result<'r> {
+        let mut response = Response::build();
+
+        if let ApiError::InternalError(_) = self {
+            // do not give away any details for internal server errors
+            // TODO: integrate with Rocket contextual logging when 0.5 is released
+            eprintln!("Internal server error: {}", self.reason());
+            response.status(http::Status::InternalServerError).ok()
+        } else {
+            // otherwise, present the error as JSON and set the status code accordingly
+            response
+                .status(self.status())
+                .header(http::ContentType::JSON)
+                .sized_body(Cursor::new(json!(self).to_string()))
+                .ok()
+        }
+    }
+}
--- a/src/routes/snippets.rs
+++ b/src/routes/snippets.rs
@@ -1,0 +1,83 @@
+use std::collections::BTreeSet;
+use std::path::PathBuf;
+
+use rocket::http::uri::Origin;
+use rocket::response::status::Created;
+use rocket::State;
+use rocket_contrib::json::JsonValue;
+use serde::Deserialize;
+
+use crate::application::Config;
+use crate::errors::ApiError;
+use crate::storage::{Changeset, Snippet, Storage};
+use crate::util::Input;
+
+#[derive(Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct NewSnippet {
+    /// Snippet title. May be omitted in the user request.
+    pub title: Option<String>,
+    /// Snippet syntax. May be omitted in the user request.
+    pub syntax: Option<String>,
+    /// Snippet content. Must be specified in the user request. Can't be empty.
+    pub content: String,
+    /// List of tags attached to the snippet. May be omitted in the user
+    /// request.
+    pub tags: Option<Vec<String>>,
+}
+
+impl NewSnippet {
+    pub fn validate(
+        self,
+        allowed_syntaxes: Option<&BTreeSet<String>>,
+    ) -> Result<Snippet, ApiError> {
+        if self.content.is_empty() {
+            return Err(ApiError::BadRequest(String::from(
+                "`content` - empty values not allowed.",
+            )));
+        }
+        if let Some(syntax) = &self.syntax {
+            if let Some(allowed_syntaxes) = &allowed_syntaxes {
+                if !allowed_syntaxes.contains(syntax) {
+                    return Err(ApiError::BadRequest(format!(
+                        "`syntax` - unallowed value {}.",
+                        syntax
+                    )));
+                }
+            }
+        }
+
+        Ok(Snippet::new(
+            self.title,
+            self.syntax,
+            vec![Changeset::new(0, self.content)],
+            self.tags.unwrap_or_default(),
+        ))
+    }
+}
+
+fn build_location(origin: &Origin, relative_path: &str) -> Result<String, ApiError> {
+    // TODO(malor): verify that this works correctly on other systems
+    let new_path = PathBuf::from(origin.path()).join(relative_path);
+    Ok(new_path
+        .to_str()
+        .ok_or_else(|| {
+            ApiError::InternalError(format!("Could not construct Location from: {:?}", new_path))
+        })?
+        .to_owned())
+}
+
+#[post("/snippets", data = "<body>")]
+pub fn create_snippet(
+    origin: &Origin,
+    config: State<Config>,
+    storage: State<Box<dyn Storage>>,
+    body: Result<Input<NewSnippet>, ApiError>,
+) -> Result<Created<JsonValue>, ApiError> {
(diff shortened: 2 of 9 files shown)


src/errors.rs
```diff
@@ -0,0 +1,83 @@
...
+
+        if let ApiError::InternalError(_) = self {
+            // do not give away any details for internal server errors
+            response.status(http::Status::InternalServerError).ok()
+        } else {
+            // otherwise, present the error as JSON and set the status code accordingly
+            response
+                .status(self.status())
+                .header(http::ContentType::JSON)
+                .sized_body(Cursor::new(json!(self).to_string()))
```
ikalnytskyi, line 79: nit: In future we probably can even analyze the content of `Accept` header and trigger corresponding serializer.

src/errors.rs
```diff
@@ -0,0 +1,83 @@
...
+    }
+}
+
+impl<'r> Responder<'r> for ApiError {
+    fn respond_to(self, _request: &Request) -> response::Result<'r> {
+        let mut response = Response::build();
+
+        if let ApiError::InternalError(_) = self {
+            // do not give away any details for internal server errors
+            response.status(http::Status::InternalServerError).ok()
```
ikalnytskyi, line 73: Is there any way to see the actual stacktrace in logs/stderr?

src/routes/snippets.rs
```diff
@@ -0,0 +1,55 @@
...
+    let NewSnippet {
+        title,
+        syntax,
+        content,
+        tags,
+    } = body.0;
+
+    if content.is_empty() {
+        return Err(ApiError::BadRequest(String::from(
+            "`content` - empty values not allowed.",
```
ikalnytskyi, line 37: nit: Should we use `:` instead of `-`? Python usually uses the former in similar situations.

tests/gabbits/create-snippet-errors.yaml
```diff
@@ -0,0 +1,59 @@
...
+        - spam
+        - eggs
+    status: 404
+
+  - name: create a new snippet (invalid JSON)
+    POST: /v1/snippets
+    request_headers:
+      content-type: application/json
+    data: spam
+    status: 400
```
ikalnytskyi, line 23: Just curious, what would be the error message? Taking into account that this is a part of normal flow (someone misspelled `content` or something), shouldn't we test the error message?

tests/gabbits/create-snippet-errors.yaml
```diff
@@ -0,0 +1,59 @@
...
+      tags:
+        - spam
+        - eggs
+    status: 400
+    response_headers:
+      content-type: application/json
+    response_json_paths:
+      $:
+        message: "`content` - empty values not allowed."
+
```
ikalnytskyi, line 42: I have two ideas for tests:

1. Test when `tags` receive a string, not list of strings.
2. Test when unexpected property is passed, e.g. `syntax1`.

tests/gabbits/create-snippet.yaml
```diff
@@ -0,0 +1,116 @@
...
+      $.syntax: python
+      $.created_at: *datetime_regex
+      $.updated_at: *datetime_regex
+      $.tags.`sorted`:
+        - eggs
+        - spam
+      $.changesets.`len`: 1
+      $.changesets[0].version: 0
+      $.changesets[0].content: print('Hello, World!')
+      $.changesets[0].created_at: *datetime_regex
```
ikalnytskyi, line 35: Nice solution! :+1:

tests/gabbits/create-snippet.yaml
```diff
@@ -0,0 +1,116 @@
...
+    response_json_paths:
+      $.id: *slug_regex
+      $.title: Hello, World!
+      $.syntax: python
+      $.created_at: *datetime_regex
+      $.updated_at: *datetime_regex
+      $.tags.`sorted`:
+        - eggs
+        - spam
+      $.changesets.`len`: 1
```
ikalnytskyi, line 32: According to our API, snippets do not expose changesets. Snippets always return the most recent changeset in its `content` property. We were planning to expose changesets via separate endpoint.

tests/gabbits/create-snippet.yaml
```diff
@@ -0,0 +1,116 @@
...
+      $.changesets[0].version: 0
+      $.changesets[0].content: print('Hello, World!')
+      $.changesets[0].created_at: *datetime_regex
+      $.changesets[0].updated_at: *datetime_regex
+
```
ikalnytskyi, line 89: I have another idea for the test:

1. Test that snippet can be created with minimum required information (i.e. just `content` as input).

tests/requirements.txt
```diff
@@ -1,2 +1,4 @@
+alembic >= 1.5.2
 gabbi >= 2.1.0
+psycopg2 >= 2.8.6
```
ikalnytskyi, line 3: psycopg2-binary to speed it up? :D

.github/workflows/tests.yml
```diff
@@ -127,4 +127,9 @@ jobs:
       - run: cargo build
       - run: python -m venv testvenv
       - run: ./testvenv/bin/python -m pip install -r tests/requirements.txt
+
+      - name: Run database migrations
+        run: |
+          ./testvenv/bin/python -m alembic.config upgrade head
```
ikalnytskyi, line 133: At some point, we will need to cleanup the database between gabbits. Shouldn't we move this to tests?

src/errors.rs
malor: Sounds good! I think that fits well into the model of Responders used in Rocket.

src/routes/snippets.rs
malor: Actually, I copied the error message from the test we had in branch v4 :) https://github.com/xsnippet/xsnippet-api/blob/version/4/tests/resources/test_snippets.py#L277-L300

I don't have a strong preference, though.

src/errors.rs
malor: It looks like Rocket is currently ongoing through a major refactoring of how logging is implemented: https://github.com/SergioBenitez/Rocket/issues/21

If you don't mind I'll just `println!()` the error to the stderr for the time being, until Rocket 0.5 is released and we can integrate with the new logging mechanism properly.

tests/gabbits/create-snippet-errors.yaml
malor: Yeah, that's a good point. At the moment, it's not pretty :( Because it's now we who produce this error message, it's not reported as all other errors that we control: instead of HTTP status + `{"message": "..."}` in the response body, you will see some text encoded as HTML.

tests/gabbits/create-snippet.yaml
malor: Ah, thanks for reminding me! I completely forgot about that.

Ok, I think this is one more reason to have a custom implementation of `Serializable` for `Snippet` rather than derive it. That way we can control the output and have the field `content` populated by the latest changeset. I'll go forward with https://github.com/xsnippet/xsnippet-api/pull/114 then and re-work this.

tests/gabbits/create-snippet.yaml
malor: Thanks! :) TBH, I don't really like that we need to test individual fields here rather than compare the whole response body with a reference, but there seems to be no way to have regexp matches in that case :(

tests/requirements.txt
malor: Oh, shoot, I forgot about that package again! :) Will do!

.github/workflows/tests.yml
malor: That's what I was thinking too. I'm glad we are on the same page about that :) Will do!

malor: I've sent #115 !

src/errors.rs
ikalnytskyi: `println!()` will work for me. Just want to make sure that we don't swallow things silently without a trace.

src/routes/snippets.rs
ikalnytskyi: It was a `nit`. I think we can think of better error messages later on.

tests/gabbits/create-snippet-errors.yaml
ikalnytskyi: Do you plan to address this later on? I wonder, should we track our planned work in issue tracker.

malor: +1 for using the issue tracker. For this one specifically, I was going to fix your comment before we merge this PR. I'll ping you when it's ready for review!

src/errors.rs
malor: It's not so much of a trace that we are used to at moment, unfortunately: it will just print the error message. I think it's ok for now, but we should check the crates from that article about error handling in Rust that you shared with me.

src/routes/snippets.rs
malor: Yeah, sounds good! I believe this one came from the validation library that we used in v4. I don't think the uses of the API should rely on the exact content of the error messages, so changing it is not a big deal.

src/errors.rs
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+
+    /// HTTP status code.
+    pub fn status(&self) -> http::Status {
+        match self {
+            ApiError::BadRequest(_) => http::Status::BadRequest,
+            ApiError::NotFound(_) => http::Status::NotFound,
+            ApiError::UnsupportedMediaType(_) => http::Status::UnsupportedMediaType,
+            _ => http::Status::InternalServerError,
```
ikalnytskyi, line 45: nit: Should we match explicitly `ApiError::InternalError(msg)` in order to avoid forgetting extending this match statement with new enums?

src/routes/snippets.rs
```diff
@@ -0,0 +1,82 @@
...
+            self.title,
+            self.syntax,
+            vec![Changeset::new(0, self.content)],
+            self.tags.unwrap_or_default(),
+        ))
+    }
+}
+
+fn build_location(origin: &Origin, relative_path: &str) -> Result<String, ApiError> {
+    let new_path = PathBuf::from(origin.path()).join(relative_path);
```
ikalnytskyi, line 60: nit: I don't think the produces string will have `/` slashes on Windows, and URLs expect `/` slashes. We will probably need to use `https://docs.rs/url/2.2.0/url/` in the future. I'm ok to stick with it for a while and check it later with tests on Windows.

src/util.rs
```diff
@@ -0,0 +1,88 @@
...
+    type Owned = String;
+    type Borrowed = str;
+
+    fn transform(r: &Request, d: Data) -> Transform<Outcome<Self::Owned, Self::Error>> {
+        let size_limit = r
+            .limits()
+            .get("max_request_size")
+            .unwrap_or(MAX_REQUEST_SIZE);
+
+        let mut buf = String::with_capacity(512);
```
ikalnytskyi, line 46: I wonder, why do we start with 512? I'd expect pretty much every JSON input will probably be bigger than 512 bytes.

src/util.rs
```diff
@@ -0,0 +1,88 @@
...
+                    if e.is_syntax() {
+                        Failure((
+                            Status::BadRequest,
+                            ApiError::BadRequest("Invalid JSON".to_string()),
+                        ))
+                    } else {
+                        Failure((Status::BadRequest, ApiError::BadRequest(e.to_string())))
+                    }
+                }
+            }
```
ikalnytskyi, line 73: If possible, I'd replace this with `Json`/`JsonValue` from `rocket_contrib` in order to avoid duplication.

ikalnytskyi (approved):
Good job! I like it. :+1:

src/errors.rs
malor: Yeah, let's do that!

src/routes/snippets.rs
malor: Good point! Let me add a TODO about that.

src/util.rs
malor: I don't have any data to back this up, but the value of 512 bytes does seem to be a bit low indeed. Let me bump this to 4k.

src/util.rs
malor: Unfortunately, because of this trait is structured, it is not straightforward :( The error type is shared between `transform()` and `from_data()`. I.e. when I set `type Error = ApiError;` above, I can no longer call the implementation of `Json::from_data()` with what our version of `transform()` has returned. To do that, we would need to create an instance of `JsonError` instead. E.g. `JsonError::Io` using some dummy `std::io::Error` (because we don't have a real one).

If you don't mind, I'll keep this as is for now and will revisit it later.

PR #113, opened 2021-01-23
Capture the logs of the xsnippet-api process started in tests

malor:
This will help with troubleshooting of obscure errors on CI.

### DIFF
--- a/tests/test_gabbits.py
+++ b/tests/test_gabbits.py
@@ -4,6 +4,7 @@ import random
 import socket
 import subprocess
 import sys
+import tempfile
 import time
 
 import gabbi.driver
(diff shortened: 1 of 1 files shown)


tests/test_gabbits.py
```diff
@@ -47,7 +48,16 @@ def start_fixture(self):
...
-        self.process = subprocess.Popen(self._launch_command, env=environ)
+        # capture stdout/stderr of xsnippet-api process to a temporary file.
+        # Alternatively, we could either connect the child process to our
+        # file descriptors (in which case, log messages from both processes
+        # would interleave), or connect them to a pipe (this is also not great,
+        # because the child process can easily fill up the pipe buffer if we do
+        # not regularly read from it in a separate thread).
+        self.application_log = tempfile.TemporaryFile()
+        self.process = subprocess.Popen(self._launch_command, env=environ,
+                                        stdout=self.application_log,
```
ikalnytskyi, line 59: If you're doing `self.application_log.read()` at the end, would it make sense to store everything in-memory immediately by using some `io.BytesIO`?

ikalnytskyi (approved):

tests/test_gabbits.py
malor: We launch a separate process, though, right? So it would not have access to the same memory. I believe we need something that has a file descriptor and can serve as an IPC mechanism. A natural choice would be using a pipe, but my concern is that pipes are backed by fixed-size buffers in the OS, and unless we keep reading from our end of the pipe here (e.g. in background thread), the writing end of the pipe in the xsnippet-api process can block as soon as it fills up that buffer.

PR #114, opened 2021-01-23
Rework Snippet serialization

malor:
In my previous attempt, I went for wrapping `chrono::DateTime` into a helper struct that allowed us to implement `Serializable` for it and automatically derive the trait for the `Snippet` struct. As it turned out, it wasn't worth it:
    
* having a wrapper around chrono::DateTime is a bit tedious and requires us to perform type conversions in different places in the code
    
* we want to have an explicit implementation of `Serializable` anyway, e.g. so that we can change what `Snippet` looks like in the API response. Specifically, we want to generate a new field - `content` - on the fly, and populate it with the content of the most recent `Changeset`


src/storage/models.rs
```diff
@@ -1,15 +1,15 @@
...
 use rand::Rng;
-use serde::Serialize;
-
-use crate::storage::DateTime;
+use serde::{ser::SerializeStruct, Serialize, Serializer};
 
 const DEFAULT_LIMIT_SIZE: usize = 20;
 const DEFAULT_SLUG_LENGTH: usize = 8;
 
+type DateTime = chrono::DateTime<chrono::Utc>;
```
ikalnytskyi, line 9: Where do we use this?

ikalnytskyi (approved):
Good job! I like it much more!

src/storage/models.rs
malor: Oh, I see how this can be confusing: for this type alias I've decided to reuse the same name that I previously used for our struct-wrapper around `chrono:DateTime`. Keeping the name allowed me to not change the lines L29 and L32 below.

PR #115, opened 2021-01-24
Make Gabbit tests run DB migrations

malor:
As briefly discussed in the comments to #112, we want to make the Gabbit tests more self-sufficient: given a database URL, they should be able to start an instance of xsnippet-api, including running the DB migrations by the means of alembic.

While we are at this, fix the warning emitted by SQLAlchemy: the name of the DB API driver has been changed from `postgres` to `postgresql`. Diesel is happy with both, so no change there.

### DIFF
--- a/tests/test_gabbits.py
+++ b/tests/test_gabbits.py
@@ -1,3 +1,4 @@
+import alembic.config
 import json
 import os
 import random
@@ -47,6 +48,9 @@ class XSnippetApi(gabbi.fixture.GabbiFixture):
 
         environ = os.environ.copy()
         environ.update(self.environ)
+
+        # run database migrations (requires ROCKET_DATABASE_URL to be set)
+        alembic.config.main(['upgrade', 'head'])
 
         # capture stdout/stderr of xsnippet-api process to a temporary file.
         # Alternatively, we could either connect the child process to our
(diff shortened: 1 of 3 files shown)


tests/test_gabbits.py
```diff
@@ -48,6 +49,9 @@ def start_fixture(self):
         environ = os.environ.copy()
         environ.update(self.environ)
 
+        # run database migrations (requires ROCKET_DATABASE_URL to be set)
+        alembic.config.main(['upgrade', 'head'])
```
ikalnytskyi, line 53: Should we also cleanup the database at `stop_fixture()`? Or you plan to add this in separate PR?

ikalnytskyi (approved):

tests/test_gabbits.py
malor: If you don't mind, I'll do it in a follow up PR. Maybe we should be more cautious and create a temporary database instead of trying to drop the data from the one that `ROCKET_DATABASE_URL` points to (same for unit tests).

ikalnytskyi: I don't mind, we can do this. But it's just something we need anyway, otherwise we won't be able to run the same tests multiple times.

malor: Yeah, agreed. I wonder if we could keep some subset of tests that we can run against a live XSnippet instance w/o access to the DB.

PR #117, opened 2021-01-31
Implement the /snippets/<id> route

ikalnytskyi:
This patch imeplements a route that retrieves the snippet by ID. With
storage layer in place, it merely a few lines of Rocket shim and dozens
lines of Gabbi tests.

### DIFF
--- a/src/routes/snippets.rs
+++ b/src/routes/snippets.rs
@@ -81,3 +81,8 @@ pub fn create_snippet(
 
     Ok(Created(location, Some(response)))
 }
+
+#[get("/snippets/<id>")]
+pub fn get_snippet(storage: State<Box<dyn Storage>>, id: String) -> Result<JsonValue, ApiError> {
+    Ok(json!(storage.get(&id)?))
+}
--- a/tests/gabbits/get-snippet-errors.yaml
+++ b/tests/gabbits/get-snippet-errors.yaml
@@ -1,0 +1,31 @@
+fixtures:
+  - XSnippetApi
+
+tests:
+  - name: create a new snippet
+    POST: /v1/snippets
+    request_headers:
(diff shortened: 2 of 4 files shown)


src/routes/snippets.rs
```diff
@@ -81,3 +81,8 @@ pub fn create_snippet(
 
     Ok(Created(location, Some(response)))
 }
+
+#[get("/snippets/<id>")]
+pub fn get_snippet(storage: State<Box<dyn Storage>>, id: String) -> Result<JsonValue, ApiError> {
```
malor, line 86: I'm not sure if we want to perform some basic validation of `id` here (e.g. to verify that it's a valid slug made only of alphanumeric characters) or just let it fail to find a snippet in the database.

tests/gabbits/get-snippet-errors.yaml
```diff
@@ -0,0 +1,31 @@
...
+    request_headers:
+      content-type: application/json
+    data:
+      title: Hello, World!
+      syntax: python
+      content: print('Hello, World!')
+    status: 201
+
+  - name: retrieve snippet (unsupported accept)
+    GET: $LOCATION
```
malor, line 16: Oh, nice! I did not know Gabbi could do that!

malor (approved):
LGTM!

src/routes/snippets.rs
ikalnytskyi: I don't think it's critical to introduce such validation at this point (because we can simply fail to lookup anything in the database). I do want to check at some point some sort of declarative validation library for Rust. I think we can benefit from it in a bunch of places. :thinking:

malor: Sounds good!

PR #118, opened 2021-01-31
Use chrono with "serde" feature enabled

malor:
This allows us not to call .to_rfc3339() explicitly and rely on chrono choosing the correct serialization format.

One difference is that the timestamps in UTC now use the Z suffix instead of encoding the HOURS:MINUTES offset (+00:00). Both are valid representations according to the standard. Account for that in the regexp that we use in tests.


ikalnytskyi (approved):
> UTC now use the Z suffix

It's actually for the better if you ask me. Most time I dealt with ISO8601, it was `Z` suffix representation in UTC. For instance, GitHub API uses `Z` suffix.

PR #119, opened 2021-01-31
Clean up construction of the Location header

malor:
We do not need anything fancy for construction of a proper Location header. The only reason for using Origin instead of hard coding a string interpolation is so that the mount prefix (/v1) is handled correctly. Other than that, we can just join the two segments: /snippets (which is fixed in the API route definition) and the new resource id.


ikalnytskyi (approved):
Great!

PR #120, opened 2021-02-06
Add a basic Dockerfile

malor:
Rust binaries are linked statically, so we can just copy the executable
to the target image. The only requirement that we have is that the image
should provide glibc and libpq.

The Dockerfile has multiple stages:

* builder: a large image that comes with rustc and Cargo installed and
  is suitable for building Rust code. We use it to compile xsnippet-api
  and prepare two object files: the xsnippet-api executable, and its
  stripped debugging symbols

* deploy (~160 MiB): an auxiliary image that can run the DB schema migrations

* base (~80 MiB): a slim target image based on Debian that will be used for
  running xsnippet-api. Provides the dynamic depedendices (glibc and libpq)
  and sets the necessary Dockerfile instructions like EXPOSE or ENV

* debug (~120 MiB): same as base, but also copies the stripped debugging symbols
  from the builder

* release (~80 MiB): same as base. Basically, it's just an alias. The only
  reason it exists is so that we can share the Dockerfile instructions between
  debug and release, and have release be the default build target

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+.gitignore
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -128,3 +128,38 @@ jobs:
       - run: python -m venv testvenv
       - run: ./testvenv/bin/python -m pip install -r tests/requirements.txt
       - run: ./testvenv/bin/python -m pytest -vv tests/
+
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Build a Docker image (deploy)
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ./Dockerfile
+          target: deploy
+          push: false
+          tags: xsnippet-api:deploy
+
+      - name: Build a Docker image (debug)
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ./Dockerfile
+          target: debug
+          push: false
+          tags: xsnippet-api:debug
+
+      - name: Build a Docker image (release)
+        uses: docker/build-push-action@v2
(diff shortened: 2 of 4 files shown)


Alexei-Kornienko:
@malor  How long it takes for you to build this image? IMHO it's not very efficient cause it has to rebuild everything from scratch...

Checkout this Dockerfile:
Basically it uses a small trick to build all deps and cache them. This way it doesn't have to recompile everything everytime code changes.
```
# Build Stage
FROM rustlang/rust:nightly-slim AS cargo-build
WORKDIR /usr/src/lunch-service

RUN apt-get update && apt-get install -y apt-utils\ 
                                         libssl-dev \
                                         pkg-config
                                         
COPY Cargo.toml Cargo.toml
RUN mkdir src/
RUN echo "fn main() {println!(\"if you see this, the build broke\")}" > src/main.rs
RUN cargo build --release
RUN cargo test
RUN rm -f target/release/deps/lunch-service*

COPY . .
RUN touch src/main.rs
RUN cargo build --release

# Final Stage
FROM ubuntu:latest
RUN apt-get update && apt-get install -y apt-utils \ 
                                         libssl-dev \
                                         pkg-config

COPY --from=cargo-build /usr/src/lunch-service/Rocket.toml /usr/local/bin/Rocket.toml
COPY --from=cargo-build /usr/src/lunch-service/target/release/lunch-service /usr/local/bin/lunch-service
CMD ["lunch-service"]

``

malor:
> @malor How long it takes for you to build this image? IMHO it's not very efficient cause it has to rebuild everything from scratch...

Yeah, at the moment it's absolutely miserable :( Thanks for the tip! I was wondering if Cargo had a built-in caching mechanism that I could reuse between builds, but apparently it's not straightforward, as you can't mount volumes during the Docker image build phase. So this trick sounds like the way to go, indeed.

malor:
We've decided to put this to https://github.com/xsnippet/xsnippet-infra instead

PR #122, opened 2021-02-07
Implement basic content type negotiation (alternative)

malor:
(an alternative implementation with content negotiation triggered directly in the Output responder)

Instead of always serialiazing responses to JSON, let's consider the
preference that a user can express via the Accept request header:

* if Accept is not passed, default to our preferred media type -
  application/json

* if Accept is passed, go through the media types specified in the
  request and pick the one with the highest preference among the ones
  that we support

* if there is no intersection between the media types that we support
  and the media types that are specified in the Accept header, give up
  and respond with `406 Not Acceptable`

At the moment, only JSON is supported.


tests/gabbits/syntaxes.yaml
```diff
@@ -12,3 +12,36 @@ tests:
...
+        - clojure
+        - json
+        - lua
+        - python
+        - rust
+
+  - name: test /syntaxes (explicitly accept any content type)
+    GET: /v1/syntaxes
+    request_headers:
+      accept: "*/*"
```
ikalnytskyi, line 33: What do you think about also testing negotiation with some 

```yaml
accept: "text/html; q=0.9, application/json"
```

and some 

```yaml
accept: "application/xml, application/json"
```

ikalnytskyi (approved):
Great!

tests/gabbits/syntaxes.yaml
malor: Good point! Added a pair of such tests to every API route.

PR #124, opened 2021-02-21
Run each DB test in a transaction

malor:
In unit tests instead of wiping out the database state between the test runs we can run each test in a database transaction that is never committed, which is equivalent to:

```
    BEGIN;

    run the test function that performs INSERTs/UPDATEs/DELETEs

    ROLLBACK;
```

This has multiple benefits:

* the very same database can now be used for local development, as its state is no longer wiped out when unit tests are executed
* tests can now be run in parallel, because each uncommitted transaction will never see modifications performed by others
* this is slightly faster, because the changes are never persisted to disk

Interestingly, updating the tests to run within a single db transaction uncovered two minor problems in the production code:

1) `SqlStorage::get()`, `SqlStorage::update()`, and `SqlStorage::list()` were using two database connections at a time. I've refactored the code to only use one, which requires delegation to a helper function that accepts a `&PgConnection`, instead of getting a connection from the pool

2) `created_at` is not guaranteed to be unique, which potentially messes up pagination if multiple snippets are created at the same time. The value of the internal id (our primary key) is now used to resolve the ties

This also allowed me to split our single giant DB test into smaller pieces.

### DIFF
--- a/src/storage/sql/mod.rs
+++ b/src/storage/sql/mod.rs
@@ -24,6 +24,26 @@ impl SqlStorage {
         Ok(Self { pool })
     }
 
+    fn get_snippet(&self, conn: &PgConnection, id: &str) -> Result<(i32, Snippet), StorageError> {
+        let result = snippets::table
+            .filter(snippets::slug.eq(id))
+            .get_result::<models::SnippetRow>(conn);
+        let snippet = match result {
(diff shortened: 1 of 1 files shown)


ikalnytskyi (approved):
Nice catch!

PR #125, opened 2021-03-06
Generate a unique id for each request and return it in the X-Request-Id header

malor:
Assigning each request a unique identifier is useful for contextual logging (will be implemented separately), and will allow us to match log entries from the application and the reverse proxy.

It does not really matter what format the identifier has; UUIDv4 seems to be a popular choice.


ikalnytskyi (approved):
Rust/Rocket is scary. Something simple as a request ID requires so much of boilerplate that is spread across a bunch of trait implementations.

malor:
> Rust/Rocket is scary. Something simple as a request ID requires so much of boilerplate that is spread across a bunch of trait implementations.

Yeah, it tries to be very generic, so some simple things might not be straightforward anymore... I don't think it is that bad, though.

PR #128, opened 2021-03-14
Add support for JWT-based authentication and authorization

malor:
Allow users to send authorized requests by passing a signed JSON Web Token value. The default auth provider is Auth0, but any provider that uses RSA for signing and validating tokens can be configured instead (which is what we do in functional tests).

XSnippet API will configure the token validator based on the values of the following environment variables"

* ROCKET_JWT_AUDIENCE (e.g. https://api.xsnippet.org)
* ROCKET_JWT_ISSUER (e.g. https://xsnippet.eu.auth0.com)
* ROCKET_JWT_JWKS_URI (e.g. https://xsnippet.eu.auth0.com/.well-known/jwks.json)

The latter specifies the location of JSON Web Key Set, which contains the _public_ keys required for validating token signatures. These keys are read once when the application starts; no additional requests to the auth provider APIs are performed. Only the asymmetric-key crypto algorihms like RS256 are supported at the moment.

As for the user and permissions, the proposed model is very basic right now:

* requests w/o the Authorization header are considered to be performed by user "Guest". This user does not have any privileges besides the ability to create and retrieve anonymous public snippets

* requests w/ the Authorization header that pass token validation are performed by user "Authenticated". An authenticated user has a unique identifier (name), and can have some additional permissions configured. Authenticated users will (when the corresponding functionality is implemented) be able to create, update, retrieve, and delete privately owned snippets
with the controlled level of visibility

* Auth providers like Auth0 allow to assign additional permissions which will be included in the signed tokens (e.g. permission "admin" can be added to for service owners).

In order to not rely on some predefined users and shared credentials, the test fixture acts as a very simple auth provider that generates a single-use RSA keypair, and a set of tokens that are passed to the Gabbi scenarios.


src/web/auth.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// snippets. May have additional permissions (e.g. if this is an admin
+    /// user).
+    Authorized {
+        name: String,
+        permissions: Vec<Permission>,
+    },
+
+    /// Anonymous user. Can create and retrieve publicly available snippets.
+    Guest,
+}
```
ikalnytskyi, line 31: I gave it a second thought and I think there's something odd with our approach. Roles normally are sets of permissions. They only make sense when we assign them to users. You know, just to make it easier to manage users by assigning them roles rather than sets of atomic permissions.

In our approach, however, a JWT token carries atomic permissions, rather than roles. That's why there's something I want to bring up for discussion.

1. The way how things currently are, I think it'd make sense to rename `Role` into `User` because it seems we will always have 2 states: anonymous and authenticated users.
2. We may end up having quite a lot of atomic permissions. Do you think it's ok to manage them on Auth0 side and send them over to API? Or should JWT tokens carry over roles, and maintain roles -> permissions mapping on API side (via some sort of configuration or something)?
3. nit: There's a slight difference between authentication and authorization.The latter implies that permissions to an action is granted. I wonder, would it make sense to use `Authenticated` instead of `Authorized`. WDYT?

src/web/auth/jwt.rs
```diff
@@ -0,0 +1,368 @@
...
+    /// Subject permissions (e.g. vec!["admin"])
+    permissions: Vec<Permission>,
+}
+
+impl Jwks {
+    /// Returns a Jwks retrieved from the location identified by the given URI.
+    pub fn from_uri(uri: &str) -> Result<Self> {
+        let load_err = Error::Configuration(format!("Can't load Jwks state from {}", uri));
+        let json = match uri.split_once("://") {
+            Some(("http", _)) | Some(("https", _)) => reqwest::blocking::get(uri)
```
ikalnytskyi, line 57: I wonder, should we trust `http://`? I mean, in case of MITM attack, the whole system is compromised. Is it not?

src/application.rs
```diff
@@ -41,6 +41,7 @@ pub fn create_app() -> Result<rocket::Rocket, Box<dyn Error>> {
         Err(err) => return Err(Box::new(err)),
     };
     let storage: Box<dyn Storage> = Box::new(SqlStorage::new(&database_url)?);
+    let token_validator: Box<dyn Validator> = Box::new(JwtValidator::from_env()?);
```
ikalnytskyi, line 44: nit: Is it possible to do something like this: `JwtValidator::from_config(app.config())`? Just to make sure our source of truth is an application config? Even more, I'd suggest putting these JWT related keys to our own `Config` struct by extracting them from `rocket.config()`, and pass that `Config` instance around.

ikalnytskyi (approved):
Overall LGTM. Good work. Just left couple of questions and comments. All could be addressed in followup PRs.

src/application.rs
```diff
@@ -3,7 +3,7 @@ use std::error::Error;
 
 use super::routes;
 use super::storage::{SqlStorage, Storage};
-use super::web::{RequestIdHeader, RequestSpan};
+use super::web::{JwtValidator, RequestIdHeader, RequestSpan, Validator};
```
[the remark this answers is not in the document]
malor: Sounds good! I thought that would be implied because the trait itself is in the `auth` module, but here we import it from `web`, so I see how this can be ambiguous.

src/application.rs
malor: Good call! Sorry, I completely forgot about our desire to support TOML-based Rocket config! :(

src/routes/snippets.rs
```diff
@@ -61,6 +61,7 @@ pub fn create_snippet(
     origin: &Origin,
     body: Result<Input<NewSnippet>, ApiError>,
     _content_type: NegotiatedContentType,
+    _role: BearerAuth,
```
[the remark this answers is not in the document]
malor: It does not. The way `BearerAuth` works is that it will try to authenticate the user if a valid token is passed in the `Authorization` header, and will resolve to `Role::Guest` if nothing is passed.

src/web/auth/jwt.rs
malor: You are absolutely right. The reason it is implemented this way at the moment is so that we can serve our own JWKS via HTTP in tests and not have to deal with the intricacies of the HTTPS setup. *However*, I've also added support for loading the JWKS state for the local filesystem, so we could use that instead and forbid HTTP completely.

Another alternative would be using conditional compilation to allow HTTP URLs in debug builds and forbid them in release builds.

malor: I've gone for the former for the time being, as it allows us to simplify the test fixture a bit. Also, it means that there is no difference between a debug and a release build of the application (in terms of behavior).

src/web/auth.rs
malor: Thank you for the feedback! These are all good points.

> The way how things currently are, I think it'd make sense to rename Role into User because it seems we will always have 2 states: anonymous and authenticated users.

Agreed. I think my choice of names for this struct was very poor and created confusion (both `Role` and `Permission::Admin`).
 
> We may end up having quite a lot of atomic permissions. Do you think it's ok to manage them on Auth0 side and send them over to API? Or should JWT tokens carry over roles, and maintain roles -> permissions mapping on API side (via some sort of configuration or something)?

That's a good question. Auth0 has notions of permissions and roles. The latter is just a way to group a set of permissions. What is communicated to us via JWT tokens is permissions. I think we have two options:

* treat permissions in Auth0 as roles in XSnippet API and then map them to permissions that are only defined in the code
* continue using permissions configured in Auth0

To be honest, I don't think we are going to have that many individual permissions, even though I'm adding one more in #130. So I suggest we stick to the latter for the time being. The downside of this approach is that we need to define permissions in two places and keep them in sync.

> nit: There's a slight difference between authentication and authorization.The latter implies that permissions to an action is granted. I wonder, would it make sense to use Authenticated instead of Authorized. WDYT?

Sure! Sounds good!

PR #129, opened 2021-03-17
Fix the tests on CI

malor:
The new release of SQLAlchemy changed the API of class URL to make it immutable. In versions >= 1.4.0 that class has a new method - `set()` - which allows to update the components of the database url (like db name) by returning a new instance of URL. That actually looks nicer than what we had before (a copy + explicit attribute assignment).


malor:
Tests are fixed in #129.

ikalnytskyi (approved):

PR #130, opened 2021-03-27
Allow to "import" snippets via POST /v1/snippets/import

malor:
Normally when new snippets are created, users are not allowed to specify the values of fields which are automatically generated,   e.g. `id`, `created_at`, and `updated_at`. This would, however, be very useful for "importing" the state of the previous incarnation of XSnippet API that is currently deployed to prod (so that we can leverage the API for this instead of INSERT'ing the values directly to the database).
    
Importing snippets should not be allowed to anyone, but only to users with the special `import` permission.

### DIFF
--- a/src/routes/snippets.rs
+++ b/src/routes/snippets.rs
@@ -1,4 +1,4 @@
-use std::collections::BTreeSet;
+use std::convert::TryFrom;
 
 use rocket::http::uri::Origin;
 use rocket::response::status::Created;
@@ -7,8 +7,19 @@ use serde::Deserialize;
 
 use crate::application::Config;
 use crate::errors::ApiError;
-use crate::storage::{Changeset, Snippet, Storage};
+use crate::storage::{Changeset, DateTime, Snippet, Storage};
 use crate::web::{BearerAuth, Input, NegotiatedContentType, Output};
+
+fn create_snippet_impl(
+    storage: &dyn Storage,
+    snippet: &Snippet,
+    base_path: &str,
+) -> Result<Created<Output<Snippet>>, ApiError> {
+    let new_snippet = storage.create(snippet)?;
+
+    let location = vec![base_path.to_string(), new_snippet.id.to_string()].join("/");
+    Ok(Created(location, Some(Output(new_snippet))))
+}
 
 #[derive(Deserialize)]
 #[serde(deny_unknown_fields)]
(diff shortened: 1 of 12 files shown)


src/routes/snippets.rs
```diff
@@ -7,9 +7,20 @@ use serde::Deserialize;
...
 use crate::web::{BearerAuth, Input, NegotiatedContentType, Output};
 
+fn create_snippet_impl(
+    storage: &dyn Storage,
+    snippet: &Snippet,
+    base_path: &str,
+) -> Result<Created<Output<Snippet>>, ApiError> {
+    let new_snippet = storage.create(snippet)?;
+
+    let location = vec![base_path.to_string(), new_snippet.id.to_string()].join("/");
```
ikalnytskyi, line 20: It's odd that Rocket doesn't provide something similar to Flask's `url_for`.

src/storage/mod.rs
```diff
@@ -3,7 +3,7 @@ mod models;
 mod sql;
 
 pub use errors::StorageError;
-pub use models::{Changeset, Direction, ListSnippetsQuery, Snippet};
+pub use models::{Changeset, DateTime, Direction, ListSnippetsQuery, Snippet};
```
ikalnytskyi, line 6: Why?

src/storage/sql/mod.rs
```diff
@@ -45,13 +45,14 @@ impl SqlStorage {
...
+        let now = chrono::Utc::now();
         let result = diesel::insert_into(snippets::table)
             .values((
                 snippets::slug.eq(&snippet.id),
                 snippets::title.eq(&snippet.title),
                 snippets::syntax.eq(&snippet.syntax),
-                snippets::created_at.eq(diesel::dsl::now),
-                snippets::updated_at.eq(diesel::dsl::now),
+                snippets::created_at.eq(snippet.created_at.unwrap_or(now)),
+                snippets::updated_at.eq(snippet.updated_at.unwrap_or(now)),
```
ikalnytskyi, line 55: So, this change essentially means now that the current timestamp is generated by the server code, not by the database (as it was before). Do we need this change because of the type requirement set by `.unwrap_or`?

src/web/auth.rs
```diff
@@ -11,16 +11,16 @@ pub use jwt::JwtValidator;
 
 #[derive(Debug, PartialEq, Deserialize)]
 pub enum Permission {
-    /// Allows superusers to perform any actions.
-    #[serde(rename(deserialize = "admin"))]
-    Admin,
+    /// Allows users to import snippets (i.e. to create new snippets and set
+    /// the values of some protected Snippet fields like `id` or `created_at`).
+    #[serde(rename(deserialize = "import"))]
+    Import,
```
ikalnytskyi, line 17: Thanks for this! :+1: I like it more this way.

ikalnytskyi (approved):
Nice work! :1st_place_medal:

src/storage/sql/mod.rs
malor: > So, this change essentially means now that the current timestamp is generated by the server code, not by the database (as it was before). 

Correct. I did not want to change this, but...

> Do we need this change because of the type requirement set by .unwrap_or?

That's actually a requirement of Rust's type system. `eq()` is a generic function that implements one of `diesel`'s SQL value traits. You can't have a *single* expression in which you conditionally pass different values to `eq()` (`diesel::dsl::now` or `snippet.created_at`), because they have different types, and thus, the compiler needs to generate/call two different versions of `eq()`. So we basically a have a choice between:

* sticking to `DateTime` values, which requires generating the timestamp on the client, and not in the database server
* having a large `if` that would repeat the same `diesel` query expression in both branches with the only difference being in the value which is passed to `eq()`

malor (commented):
Thank you! ;)

PR #137, opened 2021-06-13
Update to Rocket 0.5

malor:
We are considering ditching Rocket for axum, but, in the meantime, let's at the very least update to the latest Rocket version so that:

* we can fix all the security vulnerabilities Dependabot is complaining about

* we can use stable Rust instead of nightly

There are many backward incompatible changes, but most were due to `async` and new lifetimes sprinkled everywhere now.

One annoying bit is that we have to temporarily give up on our tracing integration: request span can no longer be created on the same thread as Tokio can move futures between threads, and the span must be created on the top-level "request" future, i.e. in Rocket itself. The code compiles, but it is disabled.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -52,8 +52,6 @@ jobs:
           - ubuntu-latest
           - windows-latest
           - macos-latest
-        rust-version:
-          - nightly
 
     runs-on: ${{ matrix.os }}
 
@@ -63,7 +61,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: ${{ matrix.rust-version }}
+          toolchain: stable
           override: true
 
       - name: Setup PostgreSQL
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -40,7 +40,7 @@ jobs:
 
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: nightly
+          toolchain: stable
           profile: minimal
           components: clippy
           override: true
@@ -57,8 +57,6 @@ jobs:
           - ubuntu-latest
           - windows-latest
           - macos-latest
-        rust-version:
-          - nightly
 
     runs-on: ${{ matrix.os }}
 
@@ -68,7 +66,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: ${{ matrix.rust-version }}
+          toolchain: stable
           override: true
 
       - name: Setup PostgreSQL
@@ -99,7 +97,7 @@ jobs:
 
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: nightly
+          toolchain: stable
           profile: minimal
           override: true
 
@@ -116,7 +114,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: nightly
+          toolchain: stable
           override: true
 
       - name: Setup PostgreSQL
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,1 +1,0 @@
-wrap_comments = true
--- a/src/application.rs
+++ b/src/application.rs
@@ -1,70 +1,77 @@
 use std::collections::BTreeSet;
-use std::error::Error;
+
+use rocket::fairing::AdHoc;
+use serde::Deserialize;
 
 use super::routes;
 use super::storage::{SqlStorage, Storage};
-use super::web::{AuthValidator, JwtValidator, RequestIdHeader, RequestSpan};
+use super::web::{AuthValidator, JwtValidator, RequestIdHeader};
 
-#[derive(Debug)]
+#[derive(Debug, Deserialize)]
 pub struct Config {
+    /// Database connection string
+    pub database_url: String,
+
     /// Keeps a set of supported syntaxes. When set, RESTful API rejects
     /// snippets with *unsupported* syntaxes. Normally this set must be
     /// kept in sync with a set of supported syntaxes in XSnippet Web in
     /// order to ensure that the web part can properly syntax-highlight
     /// snippets.
+    #[serde(default)]
     pub syntaxes: Option<BTreeSet<String>>,
 
     /// The intended recipient of the tokens (e.g. "https://api.xsnippet.org")
+    #[serde(default = "default_jwt_audience")]
     pub jwt_audience: String,
     /// The principal that issues the tokens (e.g. "https://xsnippet.eu.auth0.com/")
+    #[serde(default = "default_jwt_issuer")]
     pub jwt_issuer: String,
     /// The location of JWT Key Set with keys used to validate the tokens (e.g. "https://xsnippet.eu.auth0.com/.well-known/jwks.json")
+    #[serde(default = "default_jwks_uri")]
     pub jwt_jwks_uri: String,
 }
 
+fn default_jwt_audience() -> String {
+    "https://api.xsnippet.org".to_string()
+}
+fn default_jwt_issuer() -> String {
+    "https://xsnippet.eu.auth0.com/".to_string()
+}
+fn default_jwks_uri() -> String {
+    "https://xsnippet.eu.auth0.com/.well-known/jwks.json".to_string()
+}
+
 /// Create and return a Rocket application instance.
-///
-/// # Errors
-///
-/// Returns `Err(rocket::config::ConfigError)` if configuration supplied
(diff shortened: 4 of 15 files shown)


malor:
I have a WIP change to upgrade to diesel 2.0, which will allow us to use async/await for DB transactions.

.github/workflows/release.yml
```diff
@@ -53,7 +53,7 @@ jobs:
           - windows-latest
           - macos-latest
         rust-version:
-          - nightly
+          - stable
```
ikalnytskyi, line 56: nit: I think it'd make sense to drop `rust-version` from the matrix completely. I believe `stable` channel is set by default.

.github/workflows/tests.yml
```diff
@@ -58,7 +58,7 @@ jobs:
           - windows-latest
           - macos-latest
         rust-version:
-          - nightly
+          - stable
```
ikalnytskyi, line 61: nit: I'd either drop `rust-version` from `matrix`, or better run CI on both `stable` and `nightly` as long as `nightly` CI is ignored and is not blocking.

.rustfmt.toml
```diff
@@ -1 +0,0 @@
-wrap_comments = true
```
ikalnytskyi, line 1: Just curious, is this default now or you think it's not needed?

src/application.rs
```diff
@@ -1,70 +1,77 @@
...
     /// order to ensure that the web part can properly syntax-highlight
     /// snippets.
+    #[serde(default)]
     pub syntaxes: Option<BTreeSet<String>>,
 
     /// The intended recipient of the tokens (e.g. "https://api.xsnippet.org")
+    #[serde(default = "default_jwt_audience")]
     pub jwt_audience: String,
     /// The principal that issues the tokens (e.g. "https://xsnippet.eu.auth0.com/")
+    #[serde(default = "default_jwt_issuer")]
```
ikalnytskyi, line 27: Makes me wonder, is this even a good idea to have _defaults_ for JWT settings? Especially the one tied to Auth0? :thinking:

src/application.rs
```diff
@@ -74,11 +81,5 @@ pub fn create_app() -> Result<rocket::Rocket, Box<dyn Error>> {
         routes::syntaxes::get_syntaxes,
         routes::snippets::import_snippet,
     ];
-    Ok(app
-        .manage(config)
-        .manage(storage)
-        .manage(token_validator)
-        .attach(RequestIdHeader)
-        .attach(RequestSpan)
```
ikalnytskyi, line 82: Where did `RequestSpan` go?

src/routes/snippets.rs
```diff
@@ -155,15 +155,15 @@ fn split_marker(mut snippets: Vec<Snippet>, limit: usize) -> (Option<String>, Ve
...
-    storage: State<Box<dyn Storage>>,
-    origin: &Origin,
+pub async fn list_snippets<'h>(
+    storage: &State<Box<dyn Storage>>,
+    origin: &Origin<'_>,
     title: Option<String>,
     syntax: Option<String>,
     tag: Option<String>,
-    limit: Option<Result<PaginationLimit, ApiError>>,
+    limit: Result<PaginationLimit, rocket::form::Errors<'_>>,
```
ikalnytskyi, line 164: How come it's not an optional anymore? Is this because `limit` is always somewhere set? Is this because of `Default` trait?

src/routes/snippets.rs
```diff
@@ -282,9 +282,10 @@ pub fn get_raw_snippet(
 }
 
 #[get("/snippets/<id>", rank = 2)]
-pub fn get_snippet(
-    storage: State<Box<dyn Storage>>,
+pub async fn get_snippet(
+    storage: &State<Box<dyn Storage>>,
     id: String,
+    _content_type: &NegotiatedContentType,
```
ikalnytskyi, line 288: Just curious, why it wasn't necessary before but became necessary now?

src/web/auth/jwt.rs
```diff
@@ -53,13 +53,14 @@ struct Claims {
 
 impl Jwks {
     /// Returns a Jwks retrieved from the location identified by the given URI.
-    pub fn from_uri(uri: &str) -> Result<Self> {
-        let load_err = Error::Configuration(format!("Can't load Jwks state from {}", uri));
+    pub async fn from_uri(uri: &str) -> Result<Self> {
+        let load_err = || Error::Configuration(format!("Can't load Jwks state from {}", uri));
```
ikalnytskyi, line 57: Why is it callable and not a value?

src/web/content.rs
```diff
@@ -42,50 +38,55 @@ const PAGINATION_LIMIT_MAX: usize = 20;
...
-        let mut buf = String::with_capacity(8192);
-        match d.open().take(size_limit).read_to_string(&mut buf) {
-            Ok(_) => Borrowed(Success(buf)),
-            Err(e) => Borrowed(Failure((
-                Status::BadRequest,
-                ApiError::BadRequest(e.to_string()),
-            ))),
-        }
-    }
+            .unwrap_or_else(|| MAX_REQUEST_SIZE.bytes());
```
ikalnytskyi, line 52: Why `.unwrap_or_else()` instead of `.unwrap_or()`?

ikalnytskyi (approved):

.github/workflows/release.yml
malor: Good point!

.rustfmt.toml
malor: It's not available in `stable` yet :(

.github/workflows/tests.yml
malor: Let's drop it for now!

src/application.rs
malor: Good question! I don't remember our original thinking about this... On the one hand, I do like that we need to specify fewer config options and use "production defaults". But I also see your point about tying this to Auth0. I'll keep like this for the purpose of migration to a new Rocket release. Let's figure out what we want to do!

src/application.rs
malor: That's the annoying part: we can no longer use our implementation that was based on thread-locals, as Tokio can move futures around different threads while a request is being executed (e.g. in `await` for a DB request) :( I don't see a way to implement this properly on our end: my understanding is that Rocket needs to create a top-level tracing span. There is a feature request for that, but the work has been delayed because of the 0.5 migration. They expect it to land in 0.6.

src/routes/snippets.rs
malor: Good question!

* The form API changed, and we can no longer return our own `ApiError` from there directly
* somehow `Option` messes up route matching, and Rocket does not call `list_snippets()` anymore

But also, yes, because we also implement the `Default` trait, `Option` + `Result` are redundant, and we only need the latter.

src/web/content.rs
malor: I believe it's a `clippy` suggestion. For anything other than a constant, it prefers a closure, so that it can be evaluated lazily.

src/web/auth/jwt.rs
malor: Oh, it's the `async` thing :( I need to handle an error case twice:

* once after `reqwest::get(uri).await`
* and then right after `response.text().await`

In the sync version we could combine the two using `and_then()`, but now we need `.await` in between, which messes that up. Because I use `load_err` twice, rustc complains that the value has already been moved

```
error[E0382]: use of moved value: `load_err`
  --> src/web/auth/jwt.rs:61:47
   |
57 |         let load_err = Error::Configuration(format!("Can't load Jwks state from {}", uri));
   |             -------- move occurs because `load_err` has type `auth::Error`, which does not implement the `Copy` trait
...
60 |                 let response = reqwest::get(uri).await.map_err(|_| load_err)?;
   |                                                                --- -------- variable moved due to use in closure
   |                                                                |
   |                                                                value moved into closure here
61 |                 response.text().await.map_err(|_| load_err)?
   |                                               ^^^ -------- use occurs due to use in closure
   |                                               |
   |                                               value used here after move
```

src/routes/snippets.rs
malor: Great question!

IIRC, this is because `Responder` is not an `async` trait, and so we can no longer run request guard logic in the `Responder` implementation for `Output` (where we figure out what the format of the response should be). So we need to somehow trigger this request guard in every route. I agree that it looks weird, because the whole point of having `Output` was so that we can do it transparently, but alas :( 

Alternatively, we can make it even more explicit and pass the content type value when constructing an `Output` instance.

src/web/auth/jwt.rs
ikalnytskyi: Omg

PR #139, opened 2021-11-24
Use `ikalnytskyi/action-setup-postgres` action

ikalnytskyi:
This patch removes local `setup-postgres` github action in favor of
external `ikalnytskyi/action-setup-postgres`. The latter is essentially
the former but moves into separate repository for better code reusage
and some bugs fixed.


.github/workflows/tests.yml
```diff
@@ -71,12 +71,13 @@ jobs:
...
-      - name: Start PostgreSQL
-        uses: ./.github/actions/setup-postgres
+      - name: Setup PostgreSQL
+        uses: ikalnytskyi/action-setup-postgres@v1
+        id: postgres
 
       - name: Set ROCKET_DATABASE_URL
         run: |
-          echo "ROCKET_DATABASE_URL=postgresql://postgres:postgres@localhost/postgres" >> $GITHUB_ENV
+          echo "ROCKET_DATABASE_URL=${{ steps.postgres.outputs.connection-uri }}" >> $GITHUB_ENV
```
ikalnytskyi, line 80: **n.b.** Connection URI is now returned by the action which is great since there's no need to know default credentials now.

malor (approved):

PR #140, opened 2021-12-27
Add support for text/plain content type

malor:


malor:
We've discussed this offline and decided to proceed with #141 instead.

PR #141, opened 2021-12-27
Add a GET /snippet/:id that allows to retrieve the raw content of a snippet as plain text

malor:
This route can be used to implement the "raw" URIs in the UI w/o the need to support the corresponding nginx Lua rules.

The way we have written out content negotiation system requires us to either add support for a new content type to *all* routes (which we don't want to do at this point), or to use a new request guard to make Rocket dispatch the request to a correct function (which is a bit cumbersome but still easier).

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,12 +1,11 @@
-use std::convert::From;
-
 use rocket::http;
 use rocket::request::Request;
 use rocket::response::{self, Responder, Response};
+use rocket::Outcome;
 use serde::{ser::SerializeStruct, Serialize, Serializer};
 
 use crate::storage::StorageError;
-use crate::web::Output;
+use crate::web::{NegotiatedContentType, Output};
 
 /// All possible unsuccessful outcomes of an API request.
 ///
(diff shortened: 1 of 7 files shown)


src/errors.rs
```diff
@@ -86,12 +85,22 @@ impl<'r> Responder<'r> for ApiError {
                 .status(http::Status::InternalServerError)
                 .ok()
         } else {
-            // otherwise, present the error in the requested data format
             let http_status = self.status();
             debug!("ApiError: {:?}", self);
-            Response::build_from(Output(self).respond_to(request)?)
+
+            match request.guard::<NegotiatedContentType>() {
```
ikalnytskyi, line 91: Why it's needed all of the sudden?

ikalnytskyi (approved):

src/errors.rs
malor: Because we use `Output<T>` to serialize `ApiError`'s as well. W.o this, a 404 returned by this new route (i.e. if  you pass `Accept: text/plain; charset=utf-8`) will be partially replaced by `416 Not Acceptable`: the status will be correctly set to 404, but the body will come from https://github.com/xsnippet/xsnippet-api/blob/master/src/web/content.rs#L116-L119 (because `NegotiatedContentType` returns a failure).

I agree that it does look hacky :( But the cleanest approach I can think of is going all the way with supporting text serialization for everything (#140), or introducing a separate `MaybeTextOutput<T>` type, so that we can selectively return it in this route :(

PR #145, opened 2022-07-10
Fix master branch

ikalnytskyi:
The intention of this PR is to ensure that latest `master` can pass CI.

* It bumps versions in `Cargo.lock` because some of them contain CVEs and some others use removed APIs.
* It fixes some lint errors.
* It adds new required parameter to gabbi test generation function.


.github/workflows/tests.yml
```diff
@@ -106,7 +106,7 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: rustdoc
-          args: -- -D intra-doc-link-resolution-failure
+          args: -- -D rustdoc::broken_intra_doc_links
```
ikalnytskyi, line 109: **n.b.** It has been renamed.

src/storage/sql/models.rs
```diff
@@ -96,10 +96,10 @@ pub fn combine_rows(
                 snippet,
                 changesets_by_snippet_id
                     .remove(&snippet_id)
-                    .unwrap_or_else(Vec::new),
+                    .unwrap_or_default(),
```
ikalnytskyi, line 99: **n.b.** The linter asked for the change.

src/web/auth.rs
```diff
@@ -9,15 +9,15 @@ use serde::Deserialize;
 
 pub use jwt::JwtValidator;
 
-#[derive(Debug, PartialEq, Deserialize)]
+#[derive(Debug, PartialEq, Eq, Deserialize)]
```
ikalnytskyi, line 12: **n.b.** The linter asked for the change.

malor (approved):
Thank you! 👍

PR #152, opened 2023-11-05
Diesel 2.0

malor:


src/storage/sql/mod.rs
```diff
@@ -5,33 +5,42 @@ use std::convert::From;
...
+use diesel_async::scoped_futures::ScopedFutureExt;
+use diesel_async::{AsyncConnection, AsyncPgConnection, RunQueryDsl};
 
 use super::{errors::StorageError, Direction, ListSnippetsQuery, Snippet, Storage};
 use schema::{changesets, snippets, tags};
 
+// deadpool's default pool size is # of CPUs * 4, which might be too low. Instead,
+// let's default to PostgreSQL's max_connections value (100) and reserve a small
+// buffer for "admin" connections (like periodic database backups or local sessions).
+const MAX_POOL_SIZE: usize = 96;
```
ikalnytskyi, line 20: Since PostgreSQL `max_connections` can be changed, do you think we should make this option configurable too?

src/storage/sql/mod.rs
```diff
@@ -166,149 +181,166 @@ impl SqlStorage {
...
+    async fn create(&self, snippet: &Snippet) -> Result<Snippet, StorageError> {
+        let mut conn = self.pool.get().await?;
         conn.transaction::<_, StorageError, _>(|conn| {
-            // insert the new snippet row first to get the generated primary key
-            let snippet_id = self.insert_snippet(conn, snippet)?;
-            // insert the associated changesets
-            self.upsert_changesets(conn, snippet_id, snippet)?;
-            // insert the associated tags
-            self.upsert_tags(conn, snippet_id, snippet)?;
+            async move {
```
ikalnytskyi, line 196: What's the purpose of `async move`?

ikalnytskyi (approved):

src/storage/sql/mod.rs
malor: `async` blocks become futures. `move` works similarly to how it works in closures: values that are referenced from inside closures or `async` blocks can either be borrowed (immutably or mutably) or moved; what exactly happens depends on how we use those values (e.g. whether we call methods on them that require `self` or `&self`); `move` tells the compiler that the values must be moved and not borrowed.

All the said, that's not needed here:

* `conn` is a _reference_ passed to the closure
* `snippet` is also reference

So we don't really need `move`. I think I just copied it from the Diesel docs :P Let me clean this up!

src/storage/sql/mod.rs
malor: Yeah, let's do that as a follow up!

PR #160, opened 2024-03-18
Use gh instead of third-party GitHub actions for creating releases

malor:
Those actions started triggering various deprecation warnings, and they are generally hard to use than the CLI that we can test locally.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,29 +19,23 @@ jobs:
   create_release:
     runs-on: ubuntu-latest
     outputs:
-      upload_url: ${{ steps.create_release.outputs.upload_url }}
+      release_tag: ${{ steps.create_release.outputs.release_tag }}
     steps:
       - uses: actions/checkout@v4
 
-      - id: release_params
+      - id: create_release
         run: |
           if [[ "${GITHUB_REF}" == refs/tags/* ]]; then
-            echo "prerelease=false" >> $GITHUB_OUTPUT
             echo "release_tag=${GITHUB_REF#refs/tags/}" >> $GITHUB_OUTPUT
-            echo "title=${GITHUB_REF#refs/tags/}" >> $GITHUB_OUTPUT
+            gh release create --generate-notes "${GITHUB_REF#refs/tags/}"
           else
-            echo "prerelease=true" >> $GITHUB_OUTPUT
             echo "release_tag=nightly" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -19,29 +19,23 @@ jobs:
...
-      - id: create_release
-        uses: "marvinpinto/action-automatic-releases@latest"
-        with:
-          repo_token: ${{ secrets.GITHUB_TOKEN }}
-          prerelease: ${{ steps.release_params.outputs.prerelease }}
-          automatic_release_tag: ${{ steps.release_params.outputs.release_tag }}
-          title: ${{ steps.release_params.outputs.title }}
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.event.repository.name }}
```
ikalnytskyi, line 38: Q: We need to pass this.. em.. why? I mean, it's not `GH_REPO`, so it's unlikely picked up by `gh`, and it's not explicitly used in this shell snippet.

.github/workflows/release.yml
```diff
@@ -78,21 +72,14 @@ jobs:
...
-      - uses: actions/upload-release-asset@v1
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        with:
-          upload_url: ${{ needs.create_release.outputs.upload_url }}
-          asset_name: ${{ env.ASSET_NAME }}
-          asset_path: ${{ env.ASSET_PATH }}
-          asset_content_type: application/octet-stream
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.event.repository.name }}
```
ikalnytskyi, line 84: Ditto.

ikalnytskyi (approved):

.github/workflows/release.yml
malor: Sigh... I copied it from the example in the docs, but did not realize it wasn't for `gh` :( It should be `GH_REPO`, indeed. Good catch! :+1: 

```
> REPO=xsnippet/xsnippet-api gh release list
X No default remote repository has been set for this directory.

please run `gh repo set-default` to select a default remote repository.

> GH_REPO=xsnippet/xsnippet-api gh release list
TITLE              TYPE         TAG NAME  PUBLISHED
Development Build  Pre-release  nightly   about 2 days ago
v5.1.0             Latest       v5.1.0    about 4 months ago
v5.0.0                          v5.0.0    about 2 years ago
```

PR #170, opened 2024-11-23
Try actions/attest-build-provenance for our builds

malor:
Per https://github.blog/news-insights/product-news/introducing-artifact-attestations-now-in-public-beta/, this should allow users to verify attestations using GitHub CLI.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,6 +38,11 @@ jobs:
           GH_REPO: ${{ env.GITHUB_REPOSITORY }}
 
   build_assets:
+    permissions:
+      id-token: write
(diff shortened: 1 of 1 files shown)


ikalnytskyi (approved):
Wow! I didn't know there's an official action now. I wonder how it works? As I understand it, it doesn't upload a sigstore attestion as an artifact but rather store it somewhere on the organization level.

malor:
> Wow! I didn't know there's an official action now. I wonder how it works? As I understand it, it doesn't upload a sigstore attestion as an artifact but rather store it somewhere on the organization level.

That's my understanding as well. Let's give it a try!

PR #172, opened 2024-11-24
Cross-compile binaries for aarch64-unknown-linux-gnu

malor:
GitHub does not provide a aarch64-unknown-linux-gnu runner that we would need to compile binaries for Raspberry Pi, but we can use `cross` to build those on the amd64 Ubuntu runner.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,11 +48,20 @@ jobs:
     strategy:
       matrix:
         os:
-          - ubuntu-latest
-          - windows-latest
-          - macos-latest
+          - name: ubuntu-latest
+            cross_compile_os: linux
+            cross_compile_target: aarch64-unknown-linux-gnu
+          - name: ubuntu-latest
+            cross_compile_os:
+            cross_compile_target:
+          - name: windows-latest
+            cross_compile_os:
+            cross_compile_target:
+          - name: macos-latest
+            cross_compile_os:
+            cross_compile_target:
 
-    runs-on: ${{ matrix.os }}
+    runs-on: ${{ matrix.os.name }}
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -48,11 +48,20 @@ jobs:
...
-          - ubuntu-latest
-          - windows-latest
-          - macos-latest
-
-    runs-on: ${{ matrix.os }}
+          - name: ubuntu-latest
+            cross_compile_os: linux
+            cross_compile_target: aarch64-unknown-linux-gnu
+          - name: ubuntu-latest
+            cross_compile_os:
```
malor, line 55: Not sure if we need to specify those... If we don't, surely `${{ }}` expressions referencing those fields will fail?

.github/workflows/release.yml
```diff
@@ -66,15 +75,29 @@ jobs:
...
+        env:
+          CROSS_COMPILE_TARGET: ${{ matrix.os.cross_compile_target }}
 
       - id: build
         run: |
-          rustc --print cfg | grep = > rustc.vars
-          source rustc.vars
+          if [ -z "${CROSS_COMPILE_TARGET}" ]; then
+            # normal build targetting the CPU architecture and the OS of the host
+            rustc --print cfg | grep = > rustc.vars
```
malor, line 94: Alternatively, we can get rid of calling `rustc` to set `target_os` and `target_arch`, and specify the values for both statically in the config.

.github/workflows/release.yml
ikalnytskyi: Yeah, I don't think we need to specify this. Can't say for sure though.

.github/workflows/release.yml
```diff
@@ -48,11 +48,20 @@ jobs:
     strategy:
       matrix:
         os:
-          - ubuntu-latest
-          - windows-latest
-          - macos-latest
-
-    runs-on: ${{ matrix.os }}
+          - name: ubuntu-latest
+            cross_compile_os: linux
```
ikalnytskyi, line 52: can we somehow infer this from matrix.os? maybe using `uname` or something?

ikalnytskyi (approved):

PR #173, opened 2024-11-24
Cross-compile binaries for aarch64-unknown-linux-gnu (alternative)

malor:
GitHub does not provide a aarch64-unknown-linux-gnu runner that we would need to compile binaries for Raspberry Pi, but we can use `cross` to build those on the amd64 Ubuntu runner.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -47,10 +47,19 @@ jobs:
 
     strategy:
       matrix:
-        os:
-          - ubuntu-latest
-          - windows-latest
-          - macos-latest
+        include:
+          - target: x86_64-unknown-linux-gnu
+            os: ubuntu-latest
+            name: xsnippet-api-x86_64-linux.gz
+          - target: aarch64-unknown-linux-gnu
+            os: ubuntu-latest
+            name: xsnippet-api-aarch64-linux.gz
+          - target: x86_64-pc-windows-msvc
+            os: windows-latest
+            name: xsnippet-api-x86_64-windows.exe.7z
+          - target: aarch64-apple-darwin
+            os: macos-latest
+            name: xsnippet-api-aarch64-macos.gz
 
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -66,31 +75,34 @@ jobs:
...
 
-          pushd target/release
-          if [ "$RUNNER_OS" == "Windows" ]; then
-            export ASSET_NAME="xsnippet-api-${target_arch}-${target_os}.exe.7z"
+          export HOST="${target_arch}-${target_os}-${target_env}"
+          if [ "$HOST" = "$TARGET" ]; then
+            cargo build --release --target ${TARGET}
+          else
+            cargo install cross
+            cross build --release --target ${TARGET}
```
ikalnytskyi, line 88: nit: common denominator -- could be moved out of if-block.

malor: It's `cross build` vs `cargo build`, so there is a minor difference :(

ikalnytskyi: Ah, I see. I missed that.

ikalnytskyi (approved):

Cross.toml
```diff
@@ -0,0 +1,5 @@
+[build]
+pre-build = [
+    "dpkg --add-architecture $CROSS_DEB_ARCH",
+    "apt-get update && apt-get install --assume-yes libssl-dev:$CROSS_DEB_ARCH libpq-dev:$CROSS_DEB_ARCH"
```
ikalnytskyi, line 4: Just curious, does it mean that cross compilation doesn't work on Windows/macOS?

.github/workflows/release.yml
```diff
@@ -66,31 +75,34 @@ jobs:
...
-
       - id: build
         run: |
           rustc --print cfg | grep = > rustc.vars
           source rustc.vars
 
-          pushd target/release
-          if [ "$RUNNER_OS" == "Windows" ]; then
-            export ASSET_NAME="xsnippet-api-${target_arch}-${target_os}.exe.7z"
+          export HOST="${target_arch}-${target_os}-${target_env}"
```
ikalnytskyi, line 83: Q: I don't get it, who sets this shell variables? Or is this a part of the environment?

Cross.toml
malor: `cross` creates a Docker container behind the scenes, so, _I think_, it could work.

PR #179, opened 2024-11-26
ci: fix build error on windows

malor:
See https://github.com/sfackler/rust-openssl/issues/2149.

Apparently, in the Windows runner, commands executed in a `run` block will use the version of Perl built into MinGW instead of the one installed in the system, which breaks the openssl build because one of the expected Perl modules can't be found.

There are different ways to fix this. The one suggested in this PR is returning back to using the cargo action for "normal" (not cross compilation) builds like it used to work before. My understanding is that it helps by running cargo directly as opposed to running it from Bash.

Alternatively, we could bypass the corresponding build steps in openssl by providing locations of openssl source code via environment variables like it is suggested in the rust-openssl issue linked above.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -75,33 +75,40 @@ jobs:
       - name: Setup PostgreSQL
         uses: ikalnytskyi/action-setup-postgres@v6
 
-      - id: build
+      - id: detect_host
         run: |
-          export HOST=$(rustc -vV | grep host: | awk '{print $2}')
-          if [ "$HOST" = "$TARGET" ]; then
-            cargo build --release --target ${TARGET}
-          else
-            cargo install cross
-            cross build --release --target ${TARGET}
-          fi
+          echo "host=$(rustc -vV | grep host: | awk '{print $2}')" >> $GITHUB_OUTPUT
 
-          pushd target/${TARGET}/release
-          if [[ "$TARGET" =~ "windows" ]]; then
-            7z a $ASSET_NAME xsnippet-api.exe
+      - name: Build
+        if: ${{ steps.detect_host.outputs.host == matrix.target }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -75,33 +75,39 @@ jobs:
       - name: Setup PostgreSQL
         uses: ikalnytskyi/action-setup-postgres@v6
 
-      - id: build
+      - run: |
+          echo "HOST=$(rustc -vV | grep host: | awk '{print $2}')" >> $GITHUB_ENV
```
ikalnytskyi, line 79: nit: My personal preference would be to use outputs instead of environment variables because it's impossible to _unset_ environment variables in GitHub Actions, and because they may hypothetically clash with something else. :)

.github/workflows/release.yml
```diff
@@ -75,33 +75,39 @@ jobs:
...
           else
-            tar cvzf $ASSET_NAME xsnippet-api
+            tar cvzf ${{ matrix.name }} xsnippet-api
           fi
-          gh release upload $RELEASE_TAG $ASSET_NAME
+          gh release upload $RELEASE_TAG ${{ matrix.name }}
           popd
 
-          echo "asset_path=target/${TARGET}/release/$ASSET_NAME" >> $GITHUB_OUTPUT
+          echo "ASSET_PATH=target/${{ matrix.target }}/release/${{ matrix.name }}" >> $GITHUB_ENV
```
ikalnytskyi, line 105: ditto

ikalnytskyi (approved):

.github/workflows/release.yml
malor: Let me give it a try!

ikalnytskyi (approved):

PR #189, opened 2026-07-08
Update jsonwebtoken and other deps

malor:
* For jsonwebtoken it's a major version change, and the API is slightly different now. The surprising thing is how it no longer likes the trailing `=` characters in the Base64URL encoded values
* `cargo update` was trivial except for diesel that does not correctly specify its dependency on future-utils, so we need to add it explicitly

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,34 +3,13 @@
 version = 4
 
 [[package]]
-name = "addr2line"
-version = "0.24.2"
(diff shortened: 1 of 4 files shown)


ikalnytskyi (approved):
I hope it's AI generated :D

malor:
> I hope it's AI generated :D

Co-authored :)

